PR #2, opened 2021-09-13
Astro 1841 ag grid package

NicDominguez:
[ASTRO-1841](https://rocketcom.atlassian.net/browse/ASTRO-1841)

The goal for to create a standalone repo for our Ag Grid Astro Theme and then import it into the mono repo. 

Here is the[ standalone repo](https://github.com/RocketCommunicationsInc/ag-grid-theme)

- The current repos we have for ag-grid are an angular boiler plate app implementing the theme and a framework agnostic boiler plate for using the theme. This repo is an attempt to strip those packages down and simply provide a package that has the ag-grid css, astro custom prop variables, and the astro ag-grid theme css all in one file. 
- Assets such as fonts and images were also included in an assets sub-folder.
- The repo currently publishes under a test npm package name but should be changed to @astrouxds/ag-grid-theme when ready for release.

Including it in the monorepo should allow us to have the custom prop variables file be linked with the astro-components-stencil repo so when the stencil repo is updated with new variables so will the ag-grid theme.

-  Currently, the astro ag-grid theme package in this mono repo has a dev dependency for the stencil repo to be able to import that  dist/astro-web-components/astro-web-components.css file so our ag-grid them can use the variables. I am not positive I set this up right for lerna to handle. 

I'm looking for feedback on whether the stand alone repo is set up right, whether it publishes what we want or if something is not structured correctly to be imported into the mono repo. Further I need to know if I correctly imported into the mono-repo and the dependencies are set up properly.

The output css file from the stand alone repo was tested in the angular boiler plate application and at appears to match visually what we had before. I don't know how to better test our theme since I have never used ag-grid from scratch.


packages/ag-grid-theme/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules/
+dist
```
micahjones13, line 2: We'll want to add .DS_Store here as well, and delete the one that got pushed

nortonprojects: Once #3 is merged, there will be a .gitignore at the top level, and this one can be deleted entirely

packages/ag-grid-theme/webpack.config.js
```diff
@@ -0,0 +1,43 @@
...
+  },
+  plugins: [
+    new CopyPlugin({
+      patterns: [
+        { from: "src/fonts", to: "assets/fonts" },
+        { from: "src/icons", to: "assets/icons" },
+        { from: "src/images", to: "assets/images" },
+      ],
+    }),
+    new MiniCssExtractPlugin({
```
nortonprojects, line 20: @NicDominguez I know we talked about making this work in webpack, but have you looked at [Rollup?](https://rollupjs.org/guide/en/) It's what Stencil uses under the hood, and would likely be a lot easier.

nortonprojects (changes requested):
The general approach is good, but once #3 is merged it'll need a couple changes to work seamlessly.

Have we considered having the astro styling as a standalone package now that we're in a monorepo? That way there won't be a dependency on web components for things like this, where the components aren't used.

nortonprojects (approved):
I'm going to go ahead and approve this, and we can move stuff around when the automation stuff is all merged up

NicDominguez:
I had to do an update to all the variable names that were in the astro ag-grid theme. Once that gets merged into the stand alone repo I will re-import it into astro and we can merge.

PR #9, opened 2021-10-01
Astro 1632 popup hover

micahjones13:
## Brief Description

Adds correct hover state to rux-menu-item. 

## JIRA Link
https://rocketcom.atlassian.net/browse/ASTRO-1632

## General Notes
I believe this was the first npm i/stencil build after adding in the updated web components to the mono repo, hence the massive package-lock. 
 --color-menu-item-hover was commented out in _variables.scss so it wasn't applied. I uncommented it but it looks like it might need to be officially aliased. 

## Motivation and Context

Adds correct hover state to pop up menu.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/global/_variables.scss
```diff
@@ -138,7 +138,7 @@
     */
 
     // --color-border-menu: #2b659b;
-    // --color-menu-item-hover: #1c3f5e4d;
+    --color-menu-item-hover: #1c3f5e4d;
     // --color-segmented-button-selected: var(--color-snowflakes-dark-selected);
```
NicDominguez, line 142: This is should probably belong as an alias in the theme files. It's also one of the cases where we need to decide if we are going to apply opacity separate from a color value or include it. This is --color-snowflakes-dark-selected with an opacity of 0.3. No need to change anything now but we should remember this example for the design token discussion.

NicDominguez (approved):

packages/web-components/src/global/_variables.scss
nortonprojects: Does this fall into one of our defined colors? ie `--colorPrimaryLighten1` or something similar? Or is it a snowflake?

markacianfrani: it's `--selected`, youll want to just swap 

` --popup-menu-item-hover-background-color: var(--color-menu-item-hover);` for ` --popup-menu-item-hover-background-color: var(--selected);`

--color-menu-item-hover can be removed, it was a temp var until pop up was done in figma

micahjones13: `--selected` doesn't have an opacity tho

markacianfrani: my b, try --list-hover

micahjones13: Sweet, thanks. Now go enjoy your vacation!

nortonprojects (approved):

PR #21, opened 2021-10-04
Astro 1826 classification default

micahjones13:
## Brief Description

In SB, default classification marking read `Select a classification marking`. This changes the SB so that it defaults to `unclassfied` correctly.

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-1826

## General Notes

Removed the args declarations in SB. Because `args.classification` was undefined it was hitting the text fallback in rux-classification-banner. 
The text fallback (line 71 in .tsx) simply changes the label to be `Select a classification marking` if the classification passed in isn't valid.

## Motivation and Context

Makes the default banner in SB correct.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/stories/classification-marking.stories.mdx
+++ b/packages/web-components/src/stories/classification-marking.stories.mdx
@@ -67,7 +67,8 @@ export const WithLabel = (args) => {
 <Canvas>
     <Story
         args={{
-            label: 'customlabel',
+            classification: 'unclassified',
+            label: ' customlabel',
         }}
         name="With Label"
     >
--- a/packages/web-components/src/stories/classification-marking.stories.mdx
+++ b/packages/web-components/src/stories/classification-marking.stories.mdx
@@ -23,13 +23,20 @@ export const Default = (args) => {
         <div
             style="display: flex; flex-flow: column; justify-content: center; margin-bottom:50px; position: relative;"
         >
-            <rux-classification-marking> </rux-classification-marking>
+            <rux-classification-marking
+                classification="${args.classification}"
+                .label="${args.label}"
+                ?tag="${args.tag}"
+            >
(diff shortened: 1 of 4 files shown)


packages/web-components/src/stories/classification-marking.stories.mdx
```diff
@@ -23,12 +23,7 @@ export const Default = (args) => {
         <div
             style="display: flex; flex-flow: column; justify-content: center; margin-bottom:50px; position: relative;"
         >
-            <rux-classification-marking
```
markacianfrani, line 26: We still want these in so that the controls work

markacianfrani (changes requested):
I think we should remove the `return 'Select a Classification Marking'` entirely and either fail silently and return unclassified or throw a console warning if the classification is invalid. what do you think?

micahjones13:
> I think we should remove the `return 'Select a Classification Marking'` entirely and either fail silently and return unclassified or throw a console warning if the classification is invalid. what do you think?

Failing silently works for me, that's actually originally how I had it. We could also replace the 'select a classification marking' with 'unclassified'. Where did we land on console errors? I remember at some point we weren't going to have them.

packages/web-components/src/stories/classification-marking.stories.mdx
```diff
@@ -34,7 +34,9 @@ export const Default = (args) => {
 }
 
 <Canvas>
-    <Story name="Default">{Default.bind()}</Story>
+    <Story args={{ classification: 'unclassified' }} name="Default">
```
micahjones13, line 37: Now the control defaults to unclassified and we don't get a blank banner.

markacianfrani (approved):

PR #24, opened 2021-10-05
feat(rux-input): adds show/hide to type password

micahjones13:
## Brief Description

This adds show/hide functionality for input types of password. It also includes the `visibility` and `visibility-off` icon as the show/hide. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-2151

## General Notes

Changes made: 
- Adds togglePassword and isPasswordVisible in rux-input's State
- Adds 3 new methods: _handleTogglePassword, _handleType and _setTogglePassword
- Adds conditional rendering of rux-icon 
- adds css shadow part tag to rux-icon
- adds 2 new css classes and 1 div(icon-container) - show-password and with-label, both applied to icon-container

## Motivation and Context

This allows a user to use our rux-input with type='password' and have it give show/hide password functionality. 
We can now implement the sign-in pattern listed in this ticket: https://rocketcom.atlassian.net/browse/ASTRO-2087

## Issues and Limitations

It's important that when a password input is submitted, it's actually a type='password'. Since the only way right now to toggle show/hide is to change the input type, it's possible that the user could attempt to submit with the input type being text. This could allow the browser to add the password to it's auto-complete.
Inside of `disconnectedCallback`, if togglePassword is true, I set a check to make sure it's changed to type password. I'm not sure if this will work the way I want, so let me know if you think of a better/different way.

***EDIT***
Because of the rux-input being shadow-dom, the auto-complete shouldn't be a problem. Just in case though, I added the autocomplete attr to rux-input, on or off depending on togglePassword.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components/rux-input/rux-input.tsx
```diff
@@ -230,7 +259,15 @@ export class RuxInput implements FormFieldInterface {
                     <input
                         name={name}
                         disabled={disabled}
-                        type={type}
+                        //if type isn't password, go with type. If type is password and
+                        //the password should be visibile, then go with text. Otherwise, type.
+                        type={
```
micahjones13, line 264: I'm thinking about making this a function that returns type instead of this nested ternary since it looks messy. What do you think?

packages/web-components/src/components/rux-input/rux-input.tsx
```diff
@@ -127,17 +134,30 @@ export class RuxInput implements FormFieldInterface {
         this._handleSlotChange()
     }
 
+    //! Tried this to fix the type not changing correctly in SB. This never even fires.
+    @Watch('type')
```
micahjones13, line 138: In SB default input story, changing the type to password doesn't do anything, and this watch never fires. I've got an example using a button and js in index.html as well. What's weird is the class updates - like if you switch type to search, it will update the class, but not fire the watch.

PR #32, opened 2021-10-07
Astro 2087 sign in pattern

micahjones13:
## Brief Description

Adds a HTML/JS example of a sign in page to our SB patterns. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-2087

## General Notes

- Included in this PR is the changelog updates for the rux-input password type updates, PR here: https://github.com/RocketCommunicationsInc/astro/pull/24
- I practically copied the css from the old sign in, so it should be dead on if not just very very close. 
- The sign in example is provided in a codesandbox iframe like our form patterns are.
- I built in validation for error-text to display on rux blur, along with a validate email func that uses regex.
- In the old example, there is a SSO enabled knob that doesn't exist in this PR, because that knob is passed as a prop to the old rux-sign-in component, and this does not create a stencil version of rux-sign-in. We should discuss if that is something that we would want, however. 

## Motivation and Context

Create sign in pattern in our stencil SB.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


markacianfrani (approved):
awesome with the validation example. two quick things--

1. youll want to add type=submit on rux button.
2. the form will still submit if its not valid. I think thats fine though for this case unless you can think of a really quick way to add that. its something thats up to the dev to implement. we only need to show them how to set the error text in a real world example

PR #33, opened 2021-10-08
ASTRO-1845 - optional open and title properties, conditional render of title bar

Full-lifey:
## Brief Description

Updated the open and modalTitle properties so that they are not required. As part of this the modal title bar conditionally renders.

## JIRA Link

[ASTRO-1845](https://rocketcom.atlassian.net/browse/ASTRO-1845?atlOrigin=eyJpIjoiMGJiNWI1YTY5MjZjNDM0Zjg1YjgwZWZmZDdiZGNlYjUiLCJwIjoiaiJ9)

## General Notes

Do we want to add named and default slots to replace content?

## Issues and Limitations

Title and content is not centered. This has not changed with this update but was something I noticed:
image

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components/rux-modal/rux-modal.tsx
```diff
@@ -145,9 +145,11 @@ export class RuxModal {
...
-                                <div>{modalTitle}</div>
-                            </header>
+                            {modalTitle && (
+                                <header class="rux-modal__titlebar">
+                                    <div>{modalTitle}</div>
+                                </header>
+                            )}
                             <div class="rux-modal__content">
                                 <div class="rux-modal__message">
                                     {modalMessage}
```
Full-lifey, line 155: Specifically for the main content or `message` portion of the modal, should we have a slot and/or a prop? That would help turn it into a container for more than just text.
```suggestion
                                    <slot>{modalMessage}</slot>
```

PR #50, opened 2021-10-12
WIP Astro 2052 ag grid story

NicDominguez:
## Brief Description

Firstly, I added a new folder to the storybook sidebar to hold astro based themes. My expectation is that this folder could be used for ag-grid, material, tailwind, or any other css theme we want to offer. The AGGrid story is similar to the example provided in the previous storybook with the addition of documentation sections on installation and use of ag-grid coppied over from the ag-grid readme. 

## JIRA Link

[ASTRO-2052](https://rocketcom.atlassian.net/browse/ASTRO-2052)

## Motivation and Context

Add ag-grid documentation and storybook example to astro.

## Issues and Limitations

There is a bug that I cannot figure out. Right now the ag-grid example is rendering in the canvas tab fine. But when you switch to the docs tab the example section remains blank. If refresh on the docs page the example will render. This may be a bug caused by storybook and how those examples are rendered. The AGGrid example is unique in that the function that attaches the newGrid() to the DOM needs to be run as so `setTimeout(startAGGrid, 0)` (line 102).

## Types of changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Checklist

- [X] This PR adds or removes a Storybook story.


NicDominguez:
> I tried using a decorator and mutation observer but no dice. The ag grid docs have you listen for DOMContentLoaded instead of setTimeout, but that doesn't really make a difference. DOMContentLoaded won't get fired when you switch to the Docs tab.
> 
> I also couldn't figure out a way to just hide the example in the Docs tab either.
> 
> Would a Codesandbox be a better demo? Just to get around this issue. It feels a little overkill though.
> 
> I'm also okay with just leaving it as is vs. spending a ton of time digging into the internals of Storybook.

I think we leave it as is and see if storybook updates their api to support some kind of function call on switching tabs. As long as the canvas tab works I think that will be sufficient for most users.

PR #100, opened 2021-11-02
feat(web-components): modal rework with slots

micahjones13:
## Brief Description

Re-works rux-modal to take slots for header, message and footer.

This was a feature request from a discussion found [here.](https://github.com/RocketCommunicationsInc/astro/discussions/16#discussion-3606887)

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-2063
## General Notes
Removes all props except for `open` from `rux-modal`. Replaces these with slots for header, message and footer.
Adds `click-to-close` as a prop and no longer supports the closing of modal on an off click by default, as per our compliance.

Added tests and updated docs.

## Motivation and Context

This allows developers that use rux-modal _much_ greater control over what they can do with it. The main motivation was to allow for CRUD operations with rux-modal. 

This also allows for header, message or footer to all be ignored and not rendered, so that there can be modal variations with message, footer but no title, ect. 

## Issues and Limitations

This leads to more boilerplate by the end developer, such as: 
- Needing to focus the element they want to focus when rux-modal opens
- Filling in slots (header, content, footer) with desired elements/components
- Specific styling for content in the slots (rux-modal comes with general padding for slot containers)

All of these boilerplate items are possible and documented. 

## Types of changes

- [ ] Bug fix
- [x] New feature
- [x] Breaking change

## Checklist


markacianfrani (approved):

markacianfrani (changes requested):
Screen Shot 2021-11-19 at 9 39 21 AM

1. When using the header slot, the padding seems to be off:
```
    <rux-modal open>
      <div slot="header">ok</div>
      whatup
    </rux-modal>
```

2. We should hide the footer if there's no footer slot. 

I think it's slightly confusing that if you pass in one of the old props, it'll override your footer slot. for example, I cant do: 
```
<rux-modal modal-message="hi" open>
      <div slot="header">ok</div>
      <div slot="footer">hello</div>
    </rux-modal>
```

Not sure why I'd ever want to do that though. and I don't think there's any way to change that without it being a breaking change.

nortonprojects:
May I just say, "ok whatup" is my favorite modal of all time

micahjones13:
@markacianfrani style changes have been made. Regressions pass on the default w/ props modal, but obviously will fail with slots.

packages/web-components/src/components/rux-modal/rux-modal.tsx
```diff
@@ -50,36 +61,52 @@ export class RuxModal {
...
     }
 
-    @Watch('open')
-    validateName(isOpen: boolean) {
-        if (isOpen) {
+    connectedCallback() {
+        this._handleModalChoice = this._handleModalChoice.bind(this)
+    }
+
+    componentDidRender() {
```
micahjones13, line 101: This lifecycle method handles the auto-focusing much better. We wouldn't even need setTimeout in the _focusDefaultButton method, but it still needs it when modal is set to open by default. Otherwise, if modal is being opened by some action, the focus worked without the setTimeout.

markacianfrani (approved):
looks good! 

suggestion--should we hide the header if there's no prop or slot? what if someone doesn't want a header?

PR #103, opened 2021-11-05
Astro 2394 tests cmd

micahjones13:
## Brief Description

Adds various npm scripts to the root of astro, and updates the github workflow to use these scripts.

Added scripts: 
```
    "postinstall": "npm run bootstrap",
    "bootstrap": "lerna bootstrap",
    "build": "npm run build.web-components && npm run build.react & npm run build.angular & npm run build.astro-uxds",
    "build.tests": "npm run build.stencil && npm run build.react",
    "build.react": "npm --prefix packages/react run build",
    "build.web-components": "npm --prefix packages/web-components run build",
    "build.stencil": "npm --prefix packages/web-components run build.stencil",
    "build.astro-uxds": "npm --prefix packages/astro-uxds run build",
    "build.angular": "npm --prefix packages/angular run build",
    "test": "npm run test.web-components & npm run test.react",
    "start-server": "npm --prefix packages/web-components run ci.server",
    "start-components-test": "npm --prefix packages/web-components run test",
    "test.web-components": "start-server-and-test start-server 3333 start-components-test",
    "test.react": "npm --prefix packages/react run test",
    "test.angular": "npm --prefix packages/angular run test"
```

This also adds `start-server-and-test` and a dev dep.

This cuts the workflow down to 2 jobs - commitlint and build-and-test. This does require installing modules in both jobs, but since they run at the same time it actually makes it faster.

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-2394

## General Notes

This decreases the github action time by about half - down from around 13 min to around 5:30 min
The workflows don't actually build everything - just the necessary things in order to run tests on the packages. 
The workflow does not build:
- Storybook
- Angular
- AstroUXDS

However, the root `npm run build` will build everything. 

## Motivation and Context

Cuts down our github actions tests time by half, creates root scripts to install, build and test all packages.

## Issues and Limitations

CommitLint still fails based on old commits. 
CommitLint also gives this warning: 
```
build-and-test
Unexpected input(s) 'cache-dependency-path', valid inputs are ['always-auth', 'node-version', 'architecture', 'check-latest', 'registry-url', 'scope', 'token', 'cache', 'version']
```

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


package.json
```diff
@@ -7,9 +7,25 @@
...
+    "start-server-and-test": "^1.14.0"
   },
   "scripts": {
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "postinstall": "npm run bootstrap",
+    "bootstrap": "lerna bootstrap",
+    "build": "npm run build.web-components && npm run build.react & npm run build.angular & npm run build.astro-uxds",
+    "build.tests": "npm run build.stencil && npm run build.react",
+    "build.react": "npm --prefix packages/react run build",
```
nortonprojects, line 19: I like the idea, but I don't think it's a good idea to have a bunch of package-specific commands in our top level package, or the `npm --prefix` stuff. Just use [what Lerna has built in for filtering, etc.](https://www.npmjs.com/package/@lerna/filter-options) Like `build` could be `lerna run build`, while `build.tests` could be `lerna run --parallel --scope @(react|web-components) build`

nortonprojects (changes requested):

package.json
```diff
@@ -10,6 +10,11 @@
     "pretty-quick": "~3.1.0"
   },
   "scripts": {
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "postinstall": "npm run bootstrap",
+    "bootstrap": "lerna bootstrap",
+    "build": "lerna run build",
+    "build.tests": "lerna run --scope '{@astrouxds/astro-web-components,@astrouxds/react}' build",
+    "test": "lerna run --parallel --ignore @astrouxds/angular test"
```
nortonprojects, line 18: @micahjones13 Your previous changes worked great, so feel free to drop my changes. I mainly just moved that `start-server-and-test` command into the `web-components` package so the top level `package.json` could be as abstract as possible. 
It's approved by me, with or without my commits

nortonprojects (approved):

packages/web-components/package.json
```diff
@@ -17,7 +17,7 @@
     "scripts": {
         "build.stencil": "stencil build --docs",
         "build.storybook": "build-storybook -s ./dist",
-        "build": "npm run build.stencil && npm run build.storybook",
+        "build": "npm run build.stencil",
```
markacianfrani, line 20: just curious why was this needed? can we remove build.stencil then?

markacianfrani (approved):

package.json
micahjones13: I was trying to figure out a way to abstract the start-server-and-test and didn't even think about just moving it to web-components 😆 Thanks!

packages/web-components/package.json
micahjones13: I think it's because build.tests in the root package.json does lerna build, but we don't need storybook to build to run the tests. I think we could change the root command to be 
`lerna run --scope @astrouxds/web-components build.stencil && lerna run --scope @astrouxds/react build`
if we wanted to keep the web-components build command as it was

packages/web-components/netlify.toml
```diff
@@ -1,4 +1,4 @@
 [build]
 
 publish = "storybook-static/"
-command = "npm run build"
+command = "npm run build.stencil && npm run build.storybook"
```
nortonprojects, line 4: @markacianfrani FWIW build.storybook is still there to use with the netlify build

packages/web-components/package.json
markacianfrani: Makes sense! And I prefer it this way 100%. 9/10 times I only want to run stencil so this saves me from typing 8 characters

Issue #138, opened 2021-12-02
Rux Clock Julian Date change is not accurate for UTC


mfryder:
When working with rux-clock, I noticed that the julian date change occurs between 1400-1700 UTC rather than at 0000UTC. Here are some screenshots from the story board:
Screenshots initially taken at Dec 2 2021 UTC:
![image](https://user-images.githubusercontent.com/5786681/144468059-37ba0baf-10c6-4878-bbc1-9861bc82a3d1.png)

Story Board Screenshot:
![image](https://user-images.githubusercontent.com/5786681/144468202-ba0587cf-2247-4e72-b46f-d06c65e11c3f.png)

I checked the morning of 12/2 and found that the Julian Date for the rux-clock changes around 14-16:00 UTC:
![image](https://user-images.githubusercontent.com/5786681/144468502-6b6933c7-7e88-4450-8e32-b0bc026d3e2f.png)

![image](https://user-images.githubusercontent.com/5786681/144468544-187d0a28-4725-4437-8426-b6dec280b6b2.png)

markacianfrani:
thanks for raising this issue. we're looking into it now and will follow up here with more info

markacianfrani:
quick status update--I've identified the problem. that date calculation is (incorrectly) going off of your local system timezone, not the timezone set on the clock component. I've got a fix, but we'll need a few more days for testing since this is such a mission critical component.

mfryder:
Thanks for checking into this Mark!

markacianfrani:
Sorry for the delay here. We've just published 6.0.5 which should resolve this. Let us know if you're still having issues

PR #139, opened 2021-12-06
fix(rux-clock): date is calculated off clock tz

markacianfrani:
## Brief Description

The day of the year was being calculated based off the user's local timezone instead of the timezone that the clock was set to. 

Adds a note to the docs about the recommended format for timestamps 

## JIRA Link

ASTRO-2512

## Related Issue

[#138](https://github.com/RocketCommunicationsInc/astro/issues/138)

## Issues and Limitations

Technically, we shouldn't need to explicitly set the timezone anywhere. Clock should return the same output given any local timezone. The exception is when using strings without timezones or unix timestamps as aos/los timestamps. We may want to think about deprecating those in favor of just '1988-04-22T12:12:12.000Z'.

Snapshots were updated to use 2021 as the mock date so that we can properly check for leap years. 

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


micahjones13 (approved):
Good work!

PR #152, opened 2021-12-14
Astro 2624 modal event fires twice

micahjones13:
## Brief Description

Fixes new slot version of modal from firing ruxmodalclosed event twice when clicked to close.

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-2624

## General Notes

I'm successfully testing for the event to fire only once in the modal's e2e tests, but get this warning from cypress:
```
Cypress Warning: Cypress detected that you returned a promise in a test, but also invoked one or more cy commands inside of that promise.

The test title was:

  > Modal should fire ruxmodalclosed once if clicked to close

While this works in practice, it's often indicative of an anti-pattern. You almost never need to return both a promise and also invoke cy commands.

Cy commands themselves are already promise like, and you can likely avoid the use of the separate Promise.
```
I guessing this warning comes from:
```
        cy.get('rux-modal')
            .shadow()
            .find('.rux-modal__wrapper')
            .click('topLeft')
            .then(() => expect(count).to.equal(1))
```

## Motivation and Context

Stops ruxmodalclosed from firing twice, adds tests for it.

## Issues and Limitations

see general notes

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components/rux-modal/test/rux-modal.e2e.js
```diff
@@ -34,6 +34,24 @@ describe('Modal', () => {
             .find('.rux-modal__wrapper')
             .should('not.exist')
     })
+    it('should fire ruxmodalclosed once if clicked to close', async () => {
```
markacianfrani, line 37: here's your prob. just remove async

markacianfrani (changes requested):

markacianfrani (approved):

Issue #121, opened 2021-11-22
React boolean/removable prop issues


mrclean-oa:
I have to say this is a great library! Thank you for maintaining it. 

I have been working with the [React flavor](https://www.npmjs.com/package/@astrouxds/react) of the web component library, and have noticed some odd things around boolean/removable props. As an example, when the `RuxNotification` is in an open state, the HTML looks like this:

```html
<rux-notification status="standby" message="This is a notification banner. It won’t disappear until the user dismisses it." class="hydrated" open=""></rux-notification>
```

When it is a closed state:

```html
<rux-notification status="standby" message="This is a notification banner. It won’t disappear until the user dismisses it." class="hydrated"></rux-notification>
```

Notice that the `open` attribute is now completely missing. React does this natively with built-in elements like `input` with the `disabled` attribute. Because of this, it's most intuitive for React developers to do something like:

```jsx
<input disabled={isDisabled} />
```

But for some reason, any boolean attributes on `Rux*` components don't function the way you would expect when toggling the attribute. So I have developed this utility to help alleviate the issues as a _workaround_ but raising this issue in hopes for a more permanent fix:
```typescript
export function ruxBooleanHelper(
  condition: boolean,
  prop: string
): Record<string, boolean> {
  if (condition && prop) {
    return { [prop]: true };
  }
  return {};
}
```

I would expect to be able to toggle a boolean like `showNotification`, as an example, on the `RuxNotification` component:
```jsx
<RuxNotification open={showNotification} ... />
```
I think this needs to be solved for all `Rux*` components, since I think all boolean-related props will suffer from this. Ideally, they should function like toggling `disabled` on an input element.

Please let me know how I can help with this issue. Mark has my Email, or please mention me here.

micahjones13:
Hey @mrclean-oa, we've merged in a fix for the react boolean prop issues. It'll be available in the next publish, which will happen early next year.

PR #154, opened 2021-12-16
fix(web-components/react): apply boolean prop bindings in react correctly

micahjones13:
## Brief Description

Creates a new `react-boolean-fix.ts` file to replace some code generated by stencil's react output target. This binds boolean props to our react wrapped components the same way they are handled in react. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-2691

## Related Issue
https://github.com/RocketCommunicationsInc/astro/issues/121
## General Notes

The new boolean fix file reads and replaces a block of code with a modified boolean prop handler using regex to find the code block to replace.

## Motivation and Context

This fixes the issue with our React components not handling boolean props correctly. 
Before this fix, our react components would still register a boolean prop as true because it wasn't removing them as an attribute when they were set to false. 
Setting a boolean prop to false as 
`<RuxButton disabled={isDisabled} />` 
would result in the dom having
`<rux-button disabled='false'></rux-button>`
Which, since the disabled prop still exists, would evaluate to true.

## Issues and Limitations

Will need to keep an eye on any patches that Stencil puts out for this, as well as when React 18 launches

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


packages/react/src/react-component-lib/utils/attachProps.ts
```diff
@@ -30,21 +34,35 @@ export const attachProps = (node: HTMLElement, newProps: any, oldProps: any = {}
       } else {
         (node as any)[name] = newProps[name];
         const propType = typeof newProps[name];
-        if (propType === 'string') {
+        if (propType === "boolean") {
```
micahjones13, line 37: This is what got replaced.

packages/react/src/react-component-lib/utils/attachProps.ts
```diff
@@ -30,21 +34,35 @@ export const attachProps = (node: HTMLElement, newProps: any, oldProps: any = {}
...
+          if (newProps[name] === true) {
+            node.setAttribute(camelToDashCase(name), camelToDashCase(name));
+          } else {
+            node.removeAttribute(camelToDashCase(name));
+          }
+        } else if (propType === "string") {
```
micahjones13, line 43: This is what got replaced.

markacianfrani (approved):
whats up with the formatting noise?

Issue #157, opened 2021-12-17
readonly attribute for RuxInput


jamesathey:
Ordinary `<input>` tags can be `readonly`, which is distinct from `disabled`. A `readonly` field cannot be edited by the user, but is still submitted with the form. A `disabled` field cannot be edited by the user, but is _not_ submitted with the form.

Please add the `readonly` attribute to `<rux-input>`, and apply that attribute to the underlying `<input>` when it is present.

micahjones13:
Hey there,

Just an update that I am creating a pull request as we speak to add in the `readonly`, `spellcheck`, and `autocomplete` to RuxInput, and will comment again here when it's merged. 

Thank you for opening these issues!

micahjones13:
Alrighty, the fix has been merged and will be included in the next minor version release early next year. Thanks!

Issue #156, opened 2021-12-17
RuxInput does not support disabling autocomplete suggestions


jamesathey:
Ordinary `<input>` tags can disable autocomplete suggestions by including the attribute `autocomplete="off"`. Please add this simple attribute to `<rux-input>` and pass the attribute down to the `<input>` in the Shadow DOM.

Issue #155, opened 2021-12-17
RuxInput does not support disabling spellcheck


jamesathey:
Ordinary `<input>` tags can disable the browser's spellcheck by including the attribute `spellcheck="false"`. Please add this simple attribute to `<rux-input>` and pass the attribute down to the `<input>` in the Shadow DOM.

PR #158, opened 2021-12-17
feat(input): adds readonly, spellcheck and autocomplete props

micahjones13:
## Brief Description

Adds new props to RuxInput: readonly, spellcheck and autocomplete
Updates logic for autocomplete if type is password.

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-2696
https://rocketcom.atlassian.net/browse/ASTRO-2695

## Related Issue
https://github.com/RocketCommunicationsInc/astro/issues/155
https://github.com/RocketCommunicationsInc/astro/issues/156
https://github.com/RocketCommunicationsInc/astro/issues/157

## General Notes

Tests were added for the new props, as well as spec tests updated with new html assertions.

## Motivation and Context

Allows devs to control these attributes through the shadow dom.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components/rux-input/test/rux-input.spec.tsx
```diff
@@ -16,7 +16,7 @@ describe('rux-input', () => {
                   <slot name="label"></slot>
                 </span>
               </label>
-              <input aria-invalid="false" autocomplete="on" class="rux-input rux-input--medium" id="rux-input-1" type="text" value="">
```
micahjones13, line 19: @markacianfrani if you're wondering why this was removed, it shouldn't have been there in the first place. The logic before would change autocomplete to on if the type wasn't password, it now is not there if it's not set and is `off` if it's a pw type.

markacianfrani (approved):

PR #179, opened 2022-01-04
feat(classification-markings): tag, header and footer shadow part

micahjones13:
## Brief Description

Adds a tag, header and footer shadow part to rux-classification-marking

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-2635

## General Notes

Classification markings contains two main divs, the normal banner/tag and the footer banner (which cannot be a tag). The footer already has a shadow part named footer-banner

## Motivation and Context

shadow part effort

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-classification-marking/rux-classification-marking.tsx
+++ b/packages/web-components/src/components/rux-classification-marking/rux-classification-marking.tsx
@@ -3,9 +3,10 @@ import { Classification } from '../../common/commonTypes.module'
 import { hasSlot } from '../../utils/utils'
 
 /**
- * @part footer-banner - the footer banner
+ * @part footer-banner - the footer banner ! DEPRECATED IN FAVOR OF FOOTER !
  * @part footer - the footer banner
- * @part container - the container for the tag/banner
+ * @part tag - the container for the tag
+ * @part header - the container for the header banner
  *
  */
 @Component({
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/rux-classification-marking/rux-classification-marking.tsx
```diff
@@ -97,7 +100,7 @@ export class RuxClassificationMarking {
                             'rux-classification--banner__footer':
                                 isWrapper === true,
                         }}
-                        part="footer-banner"
+                        part="footer-banner footer"
```
micahjones13, line 103: I thought just footer would be better in the future as a part name, but I'm fine with footer-banner if we want to avoid the extra work of deprecating `footer-banner` as a part

packages/web-components/src/components/rux-classification-marking/rux-classification-marking.tsx
```diff
@@ -84,6 +86,7 @@ export class RuxClassificationMarking {
                         'rux-classification--tag': type === 'tag',
                         'rux-classification--banner': type === 'banner',
                     }}
+                    part="container"
```
micahjones13, line 89: This is only a container for the banner/tag, so maybe container isn't the best name for it?

markacianfrani: ya I agree. maybe header?

packages/web-components/src/components/rux-classification-marking/rux-classification-marking.tsx
markacianfrani: can we add "DEPRECATED" in the description for footer-banner?

markacianfrani (approved):

PR #189, opened 2022-01-06
Astro 2640 modal parts

micahjones13:
## Brief Description

Adds a container, header and content part for modal. Adds deprecation notice for wrapper part in favor of container to keep things consistent. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-2640

## General Notes

Found something interesting with modal and rux-button - rux-button has a container part on it's shadow button element, and that works for styling. But using that same part does not update the styling for the rux-modal buttons. In testing I added a part to the host of rux-button, and that would style the modal buttons but not a regular button.
This probably isn't a big deal since modal is going to be slots in the future, but thought I'd catalogue it just in case.

## Motivation and Context

Shadow part effort

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-modal/rux-modal.tsx
+++ b/packages/web-components/src/components/rux-modal/rux-modal.tsx
@@ -11,8 +11,9 @@ import {
 } from '@stencil/core'
 
 /**
- * @part wrapper - the modal wrapper overlay
- *
+ * @part wrapper - the modal wrapper overlay ! DEPRECATED IN FAVOR OF CONTAINER !
+ * @part container - the modal container
+ * @part content - the content container of the modal
  */
 @Component({
     tag: 'rux-modal',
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/rux-modal/rux-modal.tsx
```diff
@@ -11,8 +11,10 @@ import {
 } from '@stencil/core'
 
 /**
- * @part wrapper - the modal wrapper overlay
- *
+ * @part wrapper - the modal wrapper overlay ! DEPRECATED IN FAVOR OF CONTAINER !
+ * @part container - the modal container
```
micahjones13, line 15: I think all of these will be needed since they all reference something that is in the shadow dom, and this version of modal doesn't take any slots.

PR #195, opened 2022-01-11
feat(web-components): Add shadow parts to relevant components

micahjones13:
## Brief Description

Added shadow parts to relevant components.

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-2607

## General Notes

I added container shadow parts to rux-menu-item and item divider, but I'm unsure if they are completely necessary. Let me know what you think!

Also, I added label parts to components that also have label slots because those slots aren't rendering a particular component.

## Motivation and Context

Shadow part PI 

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-checkbox/rux-checkbox.tsx
+++ b/packages/web-components/src/components/rux-checkbox/rux-checkbox.tsx
@@ -19,6 +19,8 @@ let id = 0
  * @slot (default) - the label of the checkbox.
  * @part form-field - the form field wrapper container
  * @part label - the label of rux-checkbox
+ * @part form-field-message - The help text container
+ *
  */
 @Component({
     tag: 'rux-checkbox',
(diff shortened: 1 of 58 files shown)


packages/web-components/src/components/rux-checkbox/rux-checkbox.tsx
```diff
@@ -168,7 +172,7 @@ export class RuxCheckbox implements FormFieldInterface {
                             onBlur={this._onBlur}
                             ref={(el) => (this._inputEl = el)}
                         />
-                        <label htmlFor={checkboxId}>
+                        <label htmlFor={checkboxId} part="label">
```
micahjones13, line 175: @markacianfrani here is an example of adding a label part even though there is a slot associated with it. We're rending the label text, so I figured most the time a slot wouldn't be used and we should provide a part.

markacianfrani (approved):

PR #249, opened 2022-02-17
Astro 3049 prettier ignore

micahjones13:
## Brief Description

Adds the attachProps, proxies, and components.d.ts to the prettierignore. We shouldn't see those files changed when commited after a build now.

## JIRA Link
https://rocketcom.atlassian.net/browse/ASTRO-3049
## Motivation and Context

Prettier was fixing up those files after a build, so they were perpetually changed. 

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


.prettierignore
```diff
@@ -5,5 +5,9 @@ storybook-static
 *readme.md
 
 # Ignore Stencil Build files
-src/components.d.ts
+packages/web-components/src/components.d.ts
 src/globals.d.ts
```
micahjones13, line 9: I don't actually think that this file exists?

markacianfrani: oo looks like we p should have updated this during the monorepo move

markacianfrani (approved):

PR #265, opened 2022-02-23
style(status): updated size to match figma, updated bg positions

micahjones13:
## Brief Description

Changes the height and width of rux-status to be 12 px like figma (down from 1rem) 
Also updates the background positions style to match the new 12px base. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-2961

## Motivation and Context

Dev -> Design audit
## Issues and Limitations

If you change the height/width of a rux-status, you'll have some bleed through of the other statuses since we're just changing the background image position of a single image. This was the case with the 1rem based size rux-status as well, so if we want this to be changed we should move to having each status symbol be a separate svg.

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/ten-panthers-provide.md
+++ b/.changeset/ten-panthers-provide.md
@@ -1,0 +1,8 @@
+---
+"@astrouxds/astro-web-components": patch
+"@astrouxds/angular": patch
+"astro-website": patch
+"@astrouxds/react": patch
+---
+
+Changed the overall size of rux-status symbol to be 12px.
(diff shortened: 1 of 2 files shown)


micahjones13:
@markacianfrani would we consider this a breaking change? it does change the overall height/width of a status icon.

markacianfrani:
@micahjones13 technically yeah but all of this design audit stuff will just be considered patches for now. can you throw up a changeset?

markacianfrani (approved):

PR #266, opened 2022-02-24
Astro 1870 sb controls

thomasneal:
## JIRA Link

[ASTRO-1870](https://rocketcom.atlassian.net/browse/ASTRO-1870)

## Motivation and Context

Because of SB's 6.3 update, we get buttons that say "Set boolean" etc.. and we want to provide a better experience that has default values selected.

## Issues and Limitations

There is still some issues with the DateTime Utility around some options like timezone, timezone name, and locale.

## Types of changes

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/stories/tree-node.stories.mdx
```diff
@@ -51,12 +55,24 @@ export const Default = (args) => {
 
 export const Selected = (args) => {
     return html`
-<rux-tree-node ?selected="${args.selected}"> Level 1 </rux-tree-node>
+<rux-tree-node
+    ?selected="${args.selected}"
+    ?expanded="${args.expanded}"
+>
+    Level 1
+    <rux-tree-node slot="node"></rux-tree-node>
```
markacianfrani, line 63: What was the intention here adding the child slot? It might be better to just disable controls for these variants entirely because the Tree Nodes expand functionality won't work in isolation. They need to be inside a Tree component. Thinking about loud here, maybe it'd be better to have them under Tree then? similar to pop up menu items, etc.

markacianfrani (changes requested):
good stuff! just a few minor things and there's some merge conflicts it looks like. There's a few new props that have been added that might need to be pulled in here

packages/web-components/src/stories/tree-node.stories.mdx
thomasneal: This was an error when copying and pasting, thanks.

thomasneal:
@markacianfrani could you take another look? Thanks!

markacianfrani:
> @markacianfrani could you take another look? Thanks!

can you resolve the merge conflicts?

PR #311, opened 2022-03-21
fix(input): added background color to prevent inline

micahjones13:
## Brief Description

Just adds `background: var(--color-background)` to the `.rux-text-area` class in rux-input. This prevents inline styling from taking effect like we saw during the astro monday demo today

## General Notes

I'll need to check, but this might be a problem for other components as well.

## Motivation and Context

Found during Astro Monday's demo

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


markacianfrani:
@micahjones13 can you add a patch changeset and can we just apply this to all .rux-form-field classes in all the other form components as part of this?

markacianfrani (approved):

PR #315, opened 2022-03-22
Astro 3105 rux tag

micahjones13:
## Brief Description

Creates the rux-tag component.
API:
- Prop -> status -> pass, fail or unknown -> defaults to unknown
- slot -> tag-text -> Allows for tag to contain custom text
- part -> container -> the container for tag-text slot for more customizability 

Wrote e2e and unit tests as well as new story (default and all variants)
Will update regressions and open PR for backstop

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-3105

## General Notes

We will need to add a guidelines section to the docs canvas when it exists. ( I couldn't' find existing guidelines, anyway)

## Motivation and Context

Creates rux-tag which exists in design.

## Issues and Limitations

Currently only supports pass fail and unknown, and not all variants that design has.
However, since it's fully customizable, those other variants could be made by anyway who uses this component.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


.changeset/slimy-waves-design.md
```diff
@@ -0,0 +1,8 @@
+---
+"@astrouxds/angular": minor
+"@astrouxds/react": minor
+"@astrouxds/astro-web-components": minor
+"astro-website": minor
+---
+
+Created a new rux-tag component.
```
micahjones13, line 8: @markacianfrani do you want me to list API and stuff here as well?

markacianfrani: naw I think its good enough to just call it out. the docs will be more useful

packages/web-components/src/components/rux-tag/rux-tag.tsx
```diff
@@ -0,0 +1,50 @@
...
+            >
+                <div part="container">
+                    {this.status && !hasSlot(this.el, 'tag-text') ? (
+                        statusMap[this.status] ? (
+                            `${statusMap[this.status]}`
+                        ) : (
+                            'UNK'
+                        )
+                    ) : (
+                        <slot name="tag-text"></slot>
```
markacianfrani, line 44: I think this should just be the default slot so I can do `<rux-tag>whatup</rux-tag>` but I see why you did it. I dont think its blocking though

packages/web-components/src/components/rux-tag/rux-tag.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    render() {
+        return (
+            <Host
+                class={{
+                    'is-undefined': statusMap[this.status!] === undefined,
+                }}
+            >
+                <div part="container">
+                    {this.status && !hasSlot(this.el, 'tag-text') ? (
```
markacianfrani, line 37: nit: this is a little hard to read I think especially within the template. I would maybe throw this into a function and break out each if just to make it more verbose.

markacianfrani (approved):
just a few nits, looks good!

PR #338, opened 2022-04-01
Astro 2988 tree audit

micahjones13:
## Brief Description

Additions: 
- adds full borders to each rux-tree-node, as well as one selector on rux-tree to add a bottom border to the last node
- adds a new background image when selected to match design 
- added internal state `addClass`
- added new private method `_swapToLightStatus` that finds the slotted status and adds or removes the `light-theme` class depending on the below listeners. This changes the rux-status to be the light theme versions on hover as it is in Figma.
- added two `@Listeners` for `mouseover` and `mouseleave` that trigger the new `_swapToLightStatus` method
- updates the rux-tree and rux-tree-node stories to be consistent. The rux-tree-node stories specifically have a `<style>` that adds the bottom border due to the full border styling mentioned above
- Will update tests after styles check out

Figma file for reference: https://www.figma.com/file/ZtHeSWpISd8hJjN46uY9pm/Astro-UXDS-6.0---Dark-Theme?node-id=7756%3A155697

## JIRA Link

Full border -> https://rocketcom.atlassian.net/browse/ASTRO-2988

## General Notes

The addition/removal of the `light-theme` class to any slotted rux-status on hover is a lot of JS - maybe there's a better way? 

## Motivation and Context

Dev -> design audit

## Issues and Limitations

The selected svg is a fixed height - this is because if a node with children is selected and the svg doesn't have a fixed height, it will cover all it's children as well. The svg is implemented in a :host styling though, so it can be updated (although it wouldn't be super intuitive to do) 

The rux-tree-node stories have a bottom border added in a style tag. This is necessary because of how I added the full borders to every node. In order to avoid border overlaps, each node doesn't have a bottom border, but rux-tree does so it fills in the missing border on the last element. 

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/stories/tree-node.stories.mdx
```diff
@@ -55,11 +61,18 @@ export const Default = (args) => {
 
 export const Selected = (args) => {
     return html`
+<style>
+rux-tree-node{
```
markacianfrani, line 65: can we remove these? even if the result in sb looks weird without a border? nodes are only supposed to be used inside a tree and I think it's best that show our components in isolation.

also when did we add nodes inside nodes? I thought these just had single nodes.

markacianfrani (approved):

packages/web-components/src/stories/tree-node.stories.mdx
micahjones13: I did that so on the expanded example it actually showed it's arrow and stuff. I can get rid of it on the other ones, but is there any issue with keeping it for the expanded variant? Just so it looks like it would if it's expandable

micahjones13: Pushed up a new version. I removed the extra tree-node from the variants (besides expanded) and added a blurb about why the border is missing, and moved tree-node into tree on SB. lmk if you want any of that nixed

PR #353, opened 2022-04-08
feat(popup): astro-3259 popup rewrite

micahjones13:
## Brief Description

Complete re-write of rux-pop-up-menu. 
Additions:
- Rux Menu
- Floating UI 
- show/hide methods
- rux menu item changes
- tests (react tests too)
- Story update
- Trigger element is now a slot
- Pop up menu can contain anything. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-3259

## Motivation and Context

Pop up was breaking really easily - this implements floating UI which helps with the positioning and placement of the pop-up, as well as majorly simplifies the API.

## Issues and Limitations

There is a single `!important` call that I've commented on. 

## Types of changes

- [ ] Bug fix
- [x] New feature
- [x] Breaking change

## Checklist


packages/web-components/src/components/rux-pop-up-menu/rux-pop-up-menu.tsx
```diff
@@ -1,276 +1,300 @@
...
+     * @prop open - determines if the pop up is open or closed
      */
-    @Event({ eventName: 'ruxmenuwillopen' })
-    ruxMenuWillOpen!: EventEmitter<void>
-
+    @Prop({ mutable: true }) open = false
     /**
-     * Emitted when the menu is about to close
+     * @prop placement - the placement of the pop up relative to it's slotted trigger element. A list of
+     * acceptable placements can be found at [floatingui.com/docs](https://floating-ui.com/docs/computePosition#placement)
```
markacianfrani, line 53: lets pretend like we're theranos and floating ui is our seimens machine. the end user shouldn't need to know anything about floating ui. it'll make it easier to rip it out and replace it with something cool like if we ever get a native <popover>

packages/web-components/src/components/rux-pop-up-menu/rux-pop-up-menu.tsx
```diff
@@ -1,276 +1,300 @@
...
-            const caret = parseInt(getComputedStyle(this.el, ':after').height)
-            let top: number
-            let left: number
-            let caretLeft: number
-
-            const padding = 8
+    /**
+     * This returns which side the arrow is on: top, right, left or bottom.
+     * Currently using this to determine which border to bolster, but could be useful in the future.
+     * ! This does cause a 'state changed during render' stencil erorr.
```
markacianfrani, line 175: why is arrowPosition using @state? can we just make it a private var?

packages/web-components/src/components/rux-pop-up-menu/rux-pop-up-menu.tsx
```diff
@@ -1,276 +1,300 @@
...
+            )
         }
     }
 
     /**
-     * Returns 'true' if the menu is open, 'false' if it is not.
+     * @retruns Promise<boolean> - true if the pop-up is open, false if not.
      */
     @Method()
     async isOpen(): Promise<boolean> {
```
markacianfrani, line 86: can we change this to show + hide or show(boolean) ? we need an async way to be able to open and close

packages/web-components/src/components/rux-pop-up-menu/rux-pop-up-menu.tsx
```diff
@@ -1,276 +1,300 @@
...
+                        part="popup-content"
+                        ref={(el) => (this.content = el)}
+                    >
+                        <div
+                            class="rux-popup-arrow"
+                            ref={(el) => (this.arrowEl = el)}
+                            part="arrow"
+                        ></div>
+
+                        <slot></slot>
```
markacianfrani, line 295: make sure to handle onSlotchange

packages/web-components/src/components/rux-pop-up-menu/rux-pop-up-menu.tsx
```diff
@@ -1,276 +1,300 @@
...
-                <slot name="menu-end"></slot>
+            <Host aria-hidden={this.open ? 'false' : 'true'}>
+                <div class="rux-popup" part="container">
+                    <div
+                        onClick={this._handleTriggerClick}
+                        class="rux-popup__trigger"
+                        ref={(el) => (this.trigger = el)}
+                        part="trigger-container"
+                    >
+                        <slot name="trigger"></slot>
```
markacianfrani, line 267: make sure to handle onSlotchange

packages/web-components/src/components/rux-pop-up-menu/rux-pop-up-menu.tsx
micahjones13: I'm using state to trigger a re-render in order to get the correct styles applied to it. It would work as a private var if the pop-up never swapped where the arrow is, but since we're using auto-placement I need the re-render. There might be a better solution though, open to ideas

packages/web-components/src/components/rux-pop-up-menu/rux-pop-up-menu.scss
```diff
@@ -1,106 +1,53 @@
...
 }
-
-:host([open]) {
-    pointer-events: auto;
-    opacity: 1;
-    transition: opacity 0.1667s ease-in;
+.hidden {
+    //! This allows for the display to be changed with the ::part(popup-content) but still
+    //! be hidden when it needs to be. Any better ideas?
+    display: none !important;
```
micahjones13, line 13: @markacianfrani if an external dev uses the ::part(popup-content) to change the display, it'll override the .hidden and always show the pop-up. This fixes that, so when .hidden is present the display: none takes precedence. I know using !important is bad form though, so if you can think of a different way I'm all ears

markacianfrani:
noticed an oddity when you view the docs tab, the component in the iframe is missing the arrow. The placements also dont seem to be rendering but I havent dug into the story, probably something real minor

markacianfrani:
can we add support for the `strategy` attr https://floating-ui.com/docs/computePosition#strategy ? so we can have it pop out of the GSB

micahjones13:
> noticed an oddity when you view the docs tab, the component in the iframe is missing the arrow. The placements also dont seem to be rendering but I havent dug into the story, probably something real minor

I didn't see that in my local build. Was that on the deploy preview? (if so where tf is the deploy preview)

markacianfrani (changes requested):
Should be able to reproduce the disappearing caret with something like: 
Looks like an issue with the z-index?

```
<body>
        <div style=" margin: auto; width: 100px; height: 100px; background: red;">
            <rux-pop-up-menu placement="right-end" id="top">
                <rux-icon slot="trigger" icon="apps" size="small">Top</rux-icon>
                <rux-menu>
                    <rux-menu-item value="1"
                        >Pop up menu option test</rux-menu-item
                    >
                    <rux-menu-item-divider></rux-menu-item-divider>
                    <rux-menu-item value="2"
                        >Pop up menu option test</rux-menu-item
                    >
                    <rux-menu-item value="3"
                        >Pop up menu option test</rux-menu-item
                    >
                </rux-menu>
            </rux-pop-up-menu>
        </div>

        
    </body>
```

packages/web-components/src/components/rux-pop-up-menu/rux-pop-up-menu.scss
```diff
@@ -1,106 +1,53 @@
...
-    border: 8px solid transparent;
-    border-bottom: 11px solid var(--popup-menu-caret-background-color);
-    left: var(--popup-menu-caret-left, 2px);
-    top: -1.4375rem;
+    color: white;
+    padding: 12px;
+    border-radius: var(--radius-base);
+    background: var(--color-background-base-default);
+    border: 1px solid var(--color-border-interactive-muted);
+    transform-style: preserve-3d;
```
micahjones13, line 23: This and the addition of `transformZ` on the arrow is allowing for the arrow to appear at the correct z location

markacianfrani (approved):

PR #368, opened 2022-04-21
feat(container): add container component

markacianfrani:
## Brief Description

Implements the current design of Container

## JIRA Link

ASTRO-3297

## Issues and Limitations

lots of stuff here:
* No default padding on the body. 
* Too many slots. 
* Unable to reorder slots 
* No documentation

These will be addressed in a separate ticket that will change the design of the component so we can ignore them for this. This PR only aims to be a 1:1 implementation of the design in Figma. 

Other oddities:

* Table pops out of the body slot. Our table actually isn't supposed to have left/right borders either. Have a separate ticket out to address this.
* Tabs - In the tab examples, the initial content doesn't load when used in an iframe. Also separate ticket. It's an issue with the tabs component itself. 

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-container/rux-container.scss
+++ b/packages/web-components/src/components/rux-container/rux-container.scss
@@ -1,0 +1,50 @@
+:host {
+    // Internal private design tokens
+    --container-header-padding: 1rem 1.25rem; //16px 20px
+    --container-footer-padding: 1.25rem; //20px
+    --container-header-background-color: var(--color-backgrund-surface-header);
+
+    /**
+     * @prop --body-padding: the container's body padding
+     */
+    --body-padding: var(--container-footer-padding, 1.25rem);
+    display: block;
+}
+
+.hidden,
+:host([hidden]) {
+    display: none;
+}
+
+.rux-container {
+    background: var(--color-background-surface-default);
+    border: 1px solid var(--container-color-border);
+}
+
+.rux-container__header {
+    padding: var(--container-header-padding);
+    background: var(--color-background-surface-header);
+    font-family: var(--font-heading-2-font-family);
+    font-size: var(--font-heading-2-font-size);
+    font-weight: var(--font-heading-2-font-weight);
+    letter-spacing: var(--font-heading-2-letter-spacing);
+}
+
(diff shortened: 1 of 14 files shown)


packages/web-components/src/components/rux-container/rux-container.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    connectedCallback() {
+        this._handleSlotChange = this._handleSlotChange.bind(this)
+    }
+
+    private _handleSlotChange(
+        slotName: 'header' | 'tab-bar' | 'toolbar' | 'footer'
+    ) {
+        const show = hasSlot(this.el, slotName)
+        this.activeSlots = { ...this.activeSlots, [slotName]: show }
```
markacianfrani, line 40: idk this is kind of too clever and not as readable as just creating separate hasXXSlot states. But we should we removing a lot of them soon

packages/web-components/src/components/rux-container/rux-container.scss
```diff
@@ -0,0 +1,39 @@
+:host {
+    display: block;
+    --container-header-padding: 16px 20px;
```
markacianfrani, line 3: trying something new here and creating private component tokens. These are NOT css custom props and should not be documented. These will eventually be added into our design tokens. This is more or less just a way to document them before we get to that.

micahjones13 (approved):

packages/web-components/src/stories/container.stories.mdx
```diff
@@ -0,0 +1,243 @@
...
+
+<Meta
+    title="Components/Container"
+    component="rux-container"
+    argTypes={extractArgTypes('rux-container')}
+/>
+
+# Container
+
+By default, Container's have a default padding of 1.25rem (20px) set on the body. If you'd like to change this (for example, using tables or a list inside your Container), you can use the `--body-padding` CSS custom property for convenience or the `body` CSS shadow part.
```
micahjones13, line 13: Probably don't need 2 defaults here. By default, Containers have a padding of ....

PR #371, opened 2022-04-25
feat(indeterminate-progress): Astro 3075 rux-indeterminate-progress

micahjones13:
### Edit 
 Moving this to be merged into next as it's pending a re-design for smaller sizes and color in general. 

Reworks the indeterminate progress spinner to be a pure CSS solution. 
Additions: 
- New `rux-indeterminate-progress` component
- Replaces the SVG spinner on `rux-progress` with this new component
- Adds tests and stories
- Updates existing tests and stories 
- Adds a deprecation tag to rux-progress story about the future removal of the indeterminate spinner functionality of a no value `rux-progress`
- Adds a `storybook-tags` story page to hold the deprecated, beta and alpha tags. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-3075

## General Notes

I definitely need suggestions on names for the divs of the spinner. Right now I have wrapper, second third and fourth which is not intuitive. 

If rendered in a rux-progress, the size is set to 5rem (80px). If rendered as just rux-indeterminate-progress, it's 3.75rem (60px) like in Figma. 

## Motivation and Context

Updates indeterminate to match Figma, makes indeterminate it's own component

## Issues and Limitations

Still looks funky if set to less than 30px size due to the fixed borders. Using CSS parts those can be resized if necessary though, or I could look into a solution for smaller borders at smaller sizes. 

Currently using a --size css prop to set the height/width of the spinner. I know we're moving away from CSS props, plus that's an extra thing to document for dx, so if we want to just set height/width I'm ok with that

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components/rux-indeterminate-progress/rux-indeterminate-progress.scss
```diff
@@ -0,0 +1,89 @@
...
+    );
+
+    // .5's here to pervent hard stops in mask
+    --mask: radial-gradient(
+        farthest-side,
+        transparent calc((((100% - 4px) * 0.9) - 5.5px)),
+        var(--color-background-interactive-default)
+            calc((((100% - 4px) * 0.9) - 4.5px))
+    );
+    display: flex;
```
markacianfrani, line 18: can we set this to inline-block by default? I think that'll be the most common use case, what do you think?

packages/web-components/src/components/rux-indeterminate-progress/rux-indeterminate-progress.scss
```diff
@@ -0,0 +1,89 @@
+:host {
+    --size: 3.75rem;
```
markacianfrani, line 2: Would it be difficult to change this and allow standard width/height properties? If not, can we expose this as a public custom prop and add some clarification in the docs that setting width/height won't work? The other custom props here should still be private

packages/web-components/src/components/rux-indeterminate-progress/rux-indeterminate-progress.scss
```diff
@@ -0,0 +1,89 @@
...
+        farthest-side,
+        transparent calc((((100% - 4px) * 0.9) - 5.5px)),
+        var(--color-background-interactive-default)
+            calc((((100% - 4px) * 0.9) - 4.5px))
+    );
+    display: flex;
+    align-items: center;
+}
+::slotted(*) {
+    padding: 1.25rem;
```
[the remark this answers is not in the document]
markacianfrani: nvm I see, the label. It looks like the SB story at least is rendering it weird. If we're going to accept <label> in the slot, we should probably hook up its aria attributes correctly. I feel like this is outside the scope of this component though. I would have expected that the slot would have been for placing content inside the circle or something.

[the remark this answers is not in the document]
micahjones13: Yeah, I'm thinking we don't even need a slot. The figma style has a 1.25rem gap between the spinner and anything else, so I'll just go with the padding on the host and see how that looks.

packages/web-components/src/components/rux-indeterminate-progress/rux-indeterminate-progress.scss
micahjones13: I don't think so! I'll give it a shot

packages/web-components/src/components/rux-indeterminate-progress/rux-indeterminate-progress.scss
micahjones13: Agreed! just changed it.

packages/web-components/src/components/rux-indeterminate-progress/rux-indeterminate-progress.scss
```diff
@@ -0,0 +1,89 @@
...
+
+    // .5's here to pervent hard stops in mask
+    --mask: radial-gradient(
+        farthest-side,
+        transparent calc((((100% - 4px) * 0.9) - 5.5px)),
+        var(--color-background-interactive-default)
+            calc((((100% - 4px) * 0.9) - 4.5px))
+    );
+    display: inline-block;
+    padding: 1.25rem;
```
markacianfrani, line 23: can we kill this padding?

packages/web-components/src/stories/indeterminate-progress.stories.mdx
```diff
@@ -0,0 +1,56 @@
+import { extractArgTypes } from '@astrouxds/storybook-addon-docs-stencil'
+import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs'
+import { html, render } from 'lit-html'
+
+<Meta
+    title="Components/Progress/Indeterminate Progress"
```
markacianfrani, line 6: We can probably move this up one level since it's its own standalone component

packages/web-components/src/components/rux-indeterminate-progress/rux-indeterminate-progress.tsx
```diff
@@ -0,0 +1,38 @@
+import { Component, Host, h } from '@stencil/core'
+
+/**
+ * @part wrapper - The outermost div encasing rux-indeterminate-progress which is responsible for the overall size and the outermost border.
```
markacianfrani, line 4: rename to `container` for consistency with other components?

packages/web-components/src/stories/indeterminate-progress.stories.mdx
```diff
@@ -37,7 +37,7 @@ export const Default = (args) => {
 
 ## Sizing
 
-In order to re-size `rux-indeterminate-progress`, change the `--size` CSS custom property to you're desired size.
+In order to re-size `rux-indeterminate-progress`, change the `--size` CSS custom property to your desired size.
```
micahjones13, line 40: damn there goes my honorary english degree

packages/web-components/src/components/rux-indeterminate-progress/rux-indeterminate-progress.tsx
```diff
@@ -0,0 +1,80 @@
...
+    tag: 'rux-indeterminate-progress',
+    styleUrl: 'rux-indeterminate-progress.scss',
+    shadow: true,
+})
+export class RuxIndeterminateProgress {
+    @Element() el!: HTMLRuxIndeterminateProgressElement
+
+    private small: boolean = false
+
+    private _setSmall() {
```
micahjones13, line 20: @markacianfrani what do you think of the setSmall and determineMeasurement funcs? Worth my time? Any other unit measurements I should account for?

markacianfrani: Screen Shot 2022-05-06 at 9 58 20 AM
Screen Shot 2022-05-06 at 9 58 13 AM

is this the only noticeable difference between the two at 20px? I'd say just leave it out entirely for now until we have a better design story

Issue #372, opened 2022-04-25
rux-input has additional space above the input box when no label is provided


QGerrit:
This causes it to not correctly line up alongside elements which do properly fully hide the label spacing when there is no label:

`<rux-input></rux-input>`

![image](https://user-images.githubusercontent.com/97202638/165149331-c3f9462f-cc76-4a34-9e95-c2bd88d5da1f.png)

`<rux-checkbox-group></rux-checkbox-group>`

![image](https://user-images.githubusercontent.com/97202638/165149461-6a1d7ee4-8525-4163-b1f3-3dc9843b2253.png)

micahjones13:
Hey @QGerrit , a fix for this has just been merged and will be available in our release tomorrow (6.6). Thanks for bringing this up!

PR #374, opened 2022-04-26
Astro 3389 input top space

micahjones13:
## Brief Description

Removes extra margin on form elements when a label is not included. Also removes extra dom nodes (label, slot) on slider, input and textarea using conditional rendering. For other form elements, the element either didn't render extra margin already or it has specific styling on label (checkbox, radio) that inhibited the exclusion of the `<label>` element.

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-3389

## Related Issue
https://github.com/RocketCommunicationsInc/astro/issues/372
## General Notes

Checkbox and Radio both style the `<label>` in order to display anything, so I couldn't omit those when label or slot isn't being used. Instead, I added a conditional class that will override the margin if a label prop or slot isn't being used. 
In all other components, a ternary is being used to conditionally render the `<label>` or `<slot>` if they exist. 

## Motivation and Context

Removes unneeded margin on elements without labels.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/orange-onions-serve.md
+++ b/.changeset/orange-onions-serve.md
@@ -1,0 +1,8 @@
+---
+"@astrouxds/angular": patch
+"astro-website": patch
+"@astrouxds/react": patch
+"@astrouxds/astro-web-components": patch
+---
+
+Removed extra margin on form elements (checkbox, radio, slider, select, input, textarea) that didn't have a label.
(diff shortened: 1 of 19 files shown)


markacianfrani (approved):
just added a slotchange listener on checkbox

PR #377, opened 2022-04-28
fix(modal): removed fixed height on dialog

micahjones13:
## Brief Description

Removes the fixed height form the dialog element in order to allow messages of any length.

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-3294

## Related Issue

Was brought up thru email by Outside Analytics 

## General Notes

I also thought about putting a shadow part on the dialog. Thoughts? 

## Motivation and Context

Allows for longer modal message's 

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/strange-crabs-taste.md
+++ b/.changeset/strange-crabs-taste.md
@@ -1,0 +1,8 @@
+---
+"@astrouxds/astro-web-components": patch
+"@astrouxds/angular": patch
+"astro-website": patch
+"@astrouxds/react": patch
+---
+
+Removed the fix height on the dialog element in rux-modal that was preventing a long modal message.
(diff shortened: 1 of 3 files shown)


markacianfrani (approved):
shadow part sounds like a good idea. If we added one in 6.0, would it remain the same in 7.0? or would we need to deprecate it?

micahjones13:
> shadow part sounds like a good idea. If we added one in 6.0, would it remain the same in 7.0? or would we need to deprecate it?

7.0 modal is still using the dialog, so we wouldn't have to deprecate it. I'll add it real quick!

PR #385, opened 2022-05-04
docs: Add rux-button-group to cherry picking

QGerrit:
This fixes the imports for the components, and also adds `rux-button-group`, which is necessary to display the buttons with appropriate padding between them.


markacianfrani (approved):
thank you for your first contribution!

Issue #384, opened 2022-05-04
Feature Request: Add External Time prop to rux-clock


kaubster:
Add Prop _dateIn_ to rux-clock

_dateIn_ when supplied with a valid [date string or value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#syntax) overrides Date.now() standard clock.

modify updateTIme() to update _time and localDate useing _dateIn_ (when provided) otherwise default back to Date.now().

micahjones13:
Hey @kaubster, this feature has been merged in. Unfortunately the merge missed our release cadence by a few hours, so this feature will be available in the next release on the 26th.
Thanks for raising this!

PR #392, opened 2022-05-09
Astro 3444 clock datein

micahjones13:
## Brief Description

Adds a `date-in` prop to rux-clock that allows the user to set the date to be used by clock. 

Adds a snapshot test, storybook story variant as well.

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-3444

## Related Issue
https://github.com/RocketCommunicationsInc/astro/issues/384
## General Notes

Adds additional logic into the _updateTime function if `date-in` is provided. I left my comments in for now to better show my thinking behind it, I will remove them once this get's approved. 

The story is just called `With Date In` but we could call it `With Custom Date` or something. I'm fine either way

## Motivation and Context

Feat request - allow user to choose a date for clock to go by, rather than a date.now() 

## Issues and Limitations

Similar to AOS/LOS, the timestamp works best with the offset provided - if not it still uses the local timezone.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components/rux-button/readme.md
```diff
@@ -143,27 +143,6 @@ For more information about AstroUXDS usage outside of a Web Component environmen
 | `"icon"`      | the optional rux-icon                 |
 
 
-## CSS Custom Properties
```
micahjones13, line 146: I don't know why these got removed? They're still declared in the SCSS file.

markacianfrani (changes requested):
Can we get this to support a number (milliseconds since whatever, ie Date.now() - 1652122612197)  like the AOS/LOS attr do? 

Might also want to validate an actual date is passed in. with aos/los, its not so bad because it just runs once, but this will throw an error every second.

lookin good tho!

packages/web-components/src/components/rux-clock/rux-clock.tsx
```diff
@@ -119,17 +131,66 @@ export class RuxClock {
...
+            this._validateDateInUnix(this.dateIn!)
+        }
+        this._time = this.dateIn!
+        if (!this._rawTime) this._rawTime = new Date(this.dateIn!)
+        if (this._validateDateIn(this._rawTime)) {
+            this._timer = window.setInterval(() => {
+                this._updateTime()
+            }, 1000)
+        } else {
+            console.warn(
```
micahjones13, line 145: We don't have to do the console warn here, it already will produce a console error. I figured this just tells you exactly what threw the error, but I'm okay with getting rid of it

packages/web-components/src/components/rux-clock/rux-clock.tsx
```diff
@@ -119,17 +131,66 @@ export class RuxClock {
         )
     }
 
+    private _handleDateIn() {
+        if (!this.dateIn!.includes('-')) {
```
micahjones13, line 135: I'm not sure if this is the best way to determine if it's a unix date or not. Any better suggestions?

PR #411, opened 2022-05-17
Apply Prettier

jonathantneal:
## Brief Description

This updates any files that prettier will change if they are otherwise touched during development. This will reduce unrelated formatting commits in other work.

## Related Issue

This pull request is currently reliant on https://github.com/RocketCommunicationsInc/astro/pull/410

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Project configuration


markacianfrani (approved):
should we move the prettierrc out of /web-components and into the root? looks like an oversight from when we moved to monorepo

micahjones13 (approved):

jonathantneal:
Re: https://github.com/RocketCommunicationsInc/astro/pull/411#pullrequestreview-975509197 — this will be merged as-is.

After discussing the additional change with @markacianfrani and @micahjones13, we do plan to move `.prettierrc` into the project root, but we’re going to hold off the move for now. The reason we’re going to wait is that we want to be sure we’re all set for any sweeping changes this change could mean for the `next` branch. We’ll make this change in a separate pull request later on.

PR #414, opened 2022-05-23
feat(notification-banner): redesign

markacianfrani:
## Brief Description

Implements new visual design and API

## JIRA Link

ASTRO-3498

## Issues and Limitations
* Design doesn't currently have top/bottom padding so this component has a fixed height. Which means the height will not adjust with the content. For example, if you pass in a paragraph. We'll tackle that separately. 
* This entire component is due for a refresh. Requiring a parent container is cumbersome to use. This ticket is only focusing on updating the visual styles/API.
* For implementing links, I don't love the idea of needing to style the margin ourselves. 
* The only real breaking change is here removing a lot of the styling that was on `:host`
* Our status symbols are slightly larger than whats in design currently so the spacing on examples using statuses don't line up 100%

## Types of changes

- [ ] Bug fix
- [x] New feature
- [x] Breaking change

## Checklist


packages/web-components/src/components.d.ts
```diff
@@ -32386,6 +32376,10 @@ declare namespace LocalJSX {
           * If provided, the banner will automatically close after this amount of time. Accepts value either in milliseconds or seconds (which will be converted to milliseconds internally), between `2000` and `10000`, or `2` and `10`, respectively. Any number provided outside of the `2000`-`10000` range will be ignored in favor of the default 2000ms delay. <br>If `closeAfter` is not passed or if it is given an undefined or `null` value, the banner will stay open until the user closes it.
          */
         "closeAfter"?: number;
+        /**
+          * Prevents the user from dismissing the notification. Hides the `close` slot.
+         */
+        "hideClose"?: boolean;
```
markacianfrani, line 32382: spent 45 min looking for an antonym for 'dismissible' or 'closeable'. gave up

packages/web-components/src/components/rux-notification/rux-notification.tsx
```diff
@@ -41,6 +54,11 @@ export class RuxNotification {
      */
     @Prop() small: boolean = false
 
+    /**
+     * Prevents the user from dismissing the notification. Hides the `close` slot.
```
micahjones13, line 58: Should this be hides the `action` slot instead of close?

micahjones13, line 58: Should this be hides the `action` slot instead of close?

micahjones13 (approved):
One slight prop description change (hideClose) or I just read it wrong, in which case it looks good.

PR #437, opened 2022-06-08
Astro 3591 angular wrapper

micahjones13:
## Brief Description

Re-writes our angular wrapper according to the new [Stencil docs](https://stenciljs.com/docs/angular).
The issue with the angular-output-target firing events twice has been solved, so I've removed that from our stencil.config. 
The value accessors are still necessary, however some of them aren't being used - not even in the old wrapper. I'll comment where that takes place. 

I've published the package to a[ dummy repo](https://www.npmjs.com/package/@micahjones13/angular-lib) if you want to try it out, or there is a [codesandbox](https://codesandbox.io/s/peaceful-wave-l6subx?file=/src/app/app.component.html) that uses the form code that is currently on our SB as well as using Angular 14.

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-2983

## General Notes

The file structure of packages/angular has changed, but it reflects the structure in the stencil docs.
I've changed the package.json files back to be `@astrouxds/angular` and left the version at 6.8 since I'll include a changeset that will bump it eventually.

## Motivation and Context

Angular wrapper had mismatched versions of dependencies and a lot of extraneous files. This re-does the wrapper according to the documentation. It now also works with angular 13 and up. 

## Issues and Limitations

Value accessors still needed to support ng-model and forms. I used the accessor files that were being used by the old wrapper, but some of them I don't think are relevant/needed. I'll comment on those files. 

## Types of changes

- [x] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


packages/angular/projects/angular-lib/src/directives/number-value-accessor.ts
```diff
@@ -1,13 +1,13 @@
...
+import { Directive, ElementRef } from '@angular/core';
+import { NG_VALUE_ACCESSOR } from '@angular/forms';
 
-import { ValueAccessor } from "./value-accessor";
+import { ValueAccessor } from './value-accessor';
 
 @Directive({
   /* tslint:disable-next-line:directive-selector */
-  selector: "my-input[type=number]",
+  selector: 'my-input[type=number]',
```
micahjones13, line 8: @markacianfrani One of the accessor files that isn't being used, as you can see the selector is 'my-input' still. I tried a rux-input with `type=number` in an angular form (the same on on codesandbox) and it seemed to work fine, so we probably don't need this file. Leaving it here just for feedback and will delete it if necessary.

packages/angular/projects/angular-lib/src/directives/radio-value-accessor.ts
```diff
@@ -1,13 +1,13 @@
...
+import { Directive, ElementRef } from '@angular/core';
+import { NG_VALUE_ACCESSOR } from '@angular/forms';
 
-import { ValueAccessor } from "./value-accessor";
+import { ValueAccessor } from './value-accessor';
 
 @Directive({
   /* tslint:disable-next-line:directive-selector */
-  selector: "my-radio",
+  selector: 'my-radio',
```
micahjones13, line 8: Also not being used.

packages/angular/projects/angular-lib/src/directives/select-value-accessor.ts
```diff
@@ -1,13 +1,13 @@
...
+import { Directive, ElementRef } from '@angular/core';
+import { NG_VALUE_ACCESSOR } from '@angular/forms';
 
-import { ValueAccessor } from "./value-accessor";
+import { ValueAccessor } from './value-accessor';
 
 @Directive({
   /* tslint:disable-next-line:directive-selector */
-  selector: "my-range, my-radio-group",
+  selector: 'my-range, my-radio-group',
```
micahjones13, line 8: Also not being used.

PR #469, opened 2022-06-17
Astro 3591 angular wrapper

micahjones13:
## Brief Description

New angular wrapper to be used in 7.0. Merging into next.
Verified that the wrapper works in ang >=12 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-3881
## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


markacianfrani (approved):
Added the angular project to lerna so changesets would pick it up. 

This breaks our ability to just run changesets publish because of the nested directory. I made a separate ticket to try and address that and documented it in confluence. 

This PR is blocking too much so I'm okay with merging it and buying more time. We're already doing a manual release process, this just adds 2-3 extra commands, not that big of a deal IMO

PR #495, opened 2022-06-23
Update Push Button Styles

jonathantneal:


packages/web-components/src/components/rux-push-button/rux-push-button.scss
```diff
@@ -1,112 +1,175 @@
...
-    -webkit-user-select: none;
-    -moz-user-select: none;
-    -ms-user-select: none;
     user-select: none;
 
-    &[disabled] {
-        opacity: var(--disabled-opacity);
-        cursor: not-allowed;
+    /* color */
+    background-color: var(--color-background-base-default);
```
markacianfrani, line 47: where did we land on the transparent bg colors? is this something we should bring up and align with design on?

markacianfrani:
Created ASTRO-3925 to address the weird label API decision. 

The icon only variant looks way off, but it's technically outside of the scope of this ticket? Looks like there's a big chunk of CSS under `__button` that is isn't being used and the icon-only styles are in that. 

Otherwise LGTM!

markacianfrani:
closing as resolved by separate PR. keeping branch for now for ref on implementing custom units

PR #508, opened 2022-06-24
[Spacing] Radio, Radio Group

jonathantneal:


packages/web-components/src/components/rux-radio-group/rux-radio-group.tsx
```diff
@@ -144,37 +144,17 @@ export class RuxRadioGroup implements FormFieldInterface {
...
             <Host onClick={this._handleClick}>
-                <div class="rux-form-field" part="form-field">
-                    <div
-                        class={{
-                            'rux-label': true,
-                            hidden: !this.hasLabel,
-                        }}
-                        part="label"
-                    >
+                <div part="group">
```
markacianfrani, line 147: Can you update the doc block at the top of the file with this change? 

We'll also want to make sure we capture this in a changeset since it's a breaking change. For breaking changes, we roughly follow the template of:
What?: describe the change: renaming form-field part to group
Why? Why was it made? It reads better?
Migration steps: steps for the end user to follow: do a find/replace for rux-radio-group::part(form-field) -> rux-radio-group::part(group), be aware that the form-field part is used in multiple components. 

Something like that. 

Ideally, we would make a change like this in 6.0 first, add the new `group` part, mark `form-field` as deprecated, and then in 7.0, remove `form-field`. I don't think this case is that big of a deal though.

markacianfrani (changes requested):

markacianfrani:
closing as resolved by separate PR. keeping branch for now for ref on implementing custom units

PR #564, opened 2022-07-08
[Spacing] Segmented Buttons (from recent changes)

jonathantneal:
## Brief Description

Adds the missing `--button-font` variable for the font sizes. We can use the new shorthand tokens when those are available.

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-4025


markacianfrani (approved):
I still need to do some testing with the shorthand PR but lets merge this for now.

PR #581, opened 2022-07-18
Astro 3999 glossary update

micahjones13:
## Brief Description

Updates the glossary page of astrouxds with new copy/terms. Moves the sign-in pattern to be in /patterns instead of /components.

With that sign in move, I'm thinking we can get rid of the SB iframe and convert it to the 'internal.template.njk' that the other patterns are using. Thoughts? 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-3999

## Motivation and Context

Glossary update

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/astro-uxds/_content/patterns/sign-in.md
+++ b/packages/astro-uxds/_content/patterns/sign-in.md
@@ -1,6 +1,6 @@
 ---
-tags: components
-path: /components/sign-in
+tags: patterns
+path: /patterns/sign-in
 date: Last Modified
 layout: components.template.njk
 title: Sign In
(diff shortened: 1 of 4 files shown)


packages/astro-uxds/_content/patterns/sign-in.md
```diff
@@ -1,6 +1,6 @@
 ---
-tags: components
-path: /components/sign-in
+tags: patterns
+path: /patterns/sign-in
 date: Last Modified
 layout: components.template.njk
```
micahjones13, line 5: Here is were I'm thinking we drop the iframe stuff and change this layout to match the others that are under /patterns.

markacianfrani (changes requested):
sounds good to me! just make sure to add a 301 redirect from the old url to the new one. take a look at the netlify.toml config file

PR #585, opened 2022-07-19
feat(input): file type support

micahjones13:
## Brief Description

Adds support for input type=file. 
Adds tests for the new input type to make sure files can be uploaded correctly and attributes get mapped correctly.
Adds styling specific to type=file. 

This PR bumps cypress from 8 to 9.3 so that I could use the `selectFile` function in the tests. I also updated the `cypress-real-events` version to coexist better with 9.3. The tests still work as intended. 

❗ **NOTE**  - This does not include SB updates. I'll add those once I know if this is the right direction or not 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-4096

## General Notes

Instead of adding new `@Prop` or `@State`, I just used private variables which:
1. Get the file type attributes if they exist (accept, capture and multiple) https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#additional_attributes
2. Map those attributes to the shadow input

## Motivation and Context

Feat req

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components/rux-input/rux-input.scss
```diff
@@ -45,6 +45,27 @@
         cursor: pointer;
     }
 
+    input[type='file']::file-selector-button {
```
micahjones13, line 48: This works for both webkit and ff

packages/web-components/src/components/rux-input/rux-input.scss
```diff
@@ -130,6 +151,9 @@
                 url("data:image/svg+xml,%3Csvg width='40' height='40' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M35.37 26.457a15.282 15.282 0 000-21.918c-6.176-6.052-16.187-6.052-22.361 0a15.274 15.274 0 00-1.541 20.166c-.367.147-.713.37-1.014.665L.926 34.709a3.056 3.056 0 000 4.383 3.208 3.208 0 004.472 0l9.528-9.339c.352-.345.604-.753.756-1.186 6.137 3.831 14.347 3.124 19.687-2.11zM24.193 4.043c6.454 0 11.686 5.129 11.686 11.455 0 6.326-5.232 11.455-11.686 11.455-6.455 0-11.687-5.129-11.687-11.455 0-6.326 5.232-11.455 11.687-11.455z' fill='%234dacff' fill-rule='evenodd'/%3E%3C/svg%3E")
                 10px/0.975rem no-repeat;
         }
+        &--file {
```
micahjones13, line 154: Style choice I made based off of how the native looks - we might want designs input on this.

packages/web-components/src/components/rux-input/rux-input.tsx
```diff
@@ -170,13 +177,17 @@ export class RuxInput implements FormFieldInterface {
     @Watch('type')
     handleTypeChange() {
         this._setTogglePassword()
+        if (this.type === 'file') {
```
micahjones13, line 180: Just in case they are changing to a type=file on the fly, this will handle the attribute mapping.

PR #593, opened 2022-07-21
fix(rux-tabs): can no longer change shadow dom typography styles

markacianfrani:
## Brief Description

Part of https://github.com/RocketCommunicationsInc/astro/pull/589. Breaking into it's own PR because its pretty chonky and required some API changes. 

Tabs relied entirely on host styles so I needed to create an internal class and an explicit `small` prop/attr. 

There is one major breaking change in that rux-tab by itself is now large by default. It was previously small for some reason, which didn't make sense because rux-tabs is large by default. 

## Issues and Limitations

Our dev implementation of design is actually way off after looking at Figma closer. This should hopefully get smoothed out in the spacing effort, I was only concerned with getting our VRT to pass here. 

## Types of changes

- [x] Bug fix
- [x] New feature
- [x] Breaking change

## Checklist

### DIFF
--- a/.changeset/loud-planes-accept.md
+++ b/.changeset/loud-planes-accept.md
@@ -1,0 +1,9 @@
+---
+"angular-workspace": minor
+"@astrouxds/angular": minor
+"astro-website": minor
+"@astrouxds/react": minor
+"@astrouxds/astro-web-components": minor
+---
+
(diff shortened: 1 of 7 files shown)


micahjones13 (approved):
LGTM. idk who put all styles on host and made the default small without thinking twice, shame on them 😅 😨

PR #610, opened 2022-08-01
Fix typos in cherry picking for tabs

QGerrit:
Just fixes some typos I noticed in the documentation.


micahjones13 (approved):
Good catch, thanks for the PR @QGerrit!

PR #620, opened 2022-08-05
fix(timeline): editing time events

markacianfrani:
## Brief Description

Steps to reproduce:

Change a Time Region's start or end attribute after it's already been rendered. It should update accordingly.

## JIRA Link

ASTRO-4314

## Issues and Limitations

The time region position is calculated from it's parent Track. The problem is that Track doesn't know when a Time Region's start or end date changes. 

Went with an event,  but maybe a mutation observer would've been the better choice? 

Also, tried to write a test for this and gave up.
```
    it('updates the position of an event when its start or end date is changed', () => {
		cy.get('rux-time-region')
		.invoke('prop', 'start', '2022-08-08T01:00')
		.should('have.css', 'grid-area', '1 / 62 / auto / 2882')
    })
```

Passes without any code changes. 

Same test in Playwright:
```
test('homepage has Playwright in title and get started link linking to the intro page', async ({ page }) => {
  await page.goto('http://localhost:3333/components/rux-timeline/test');
  const timeline = await page.locator('rux-time-region')
  await timeline.evaluate(node => node.setAttribute('start', '2022-08-08T01:00'))
  await expect(timeline).toHaveCSS('grid-area', '1 / 62 / auto / 2882')
});
```
Works like a normal person

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-timeline/rux-track/rux-track.tsx
+++ b/packages/web-components/src/components/rux-timeline/rux-track/rux-track.tsx
@@ -1,4 +1,4 @@
-import { Element, Component, Prop, Host, h, Watch } from '@stencil/core'
+import { Element, Listen, Component, Prop, Host, h, Watch } from '@stencil/core'
 import { differenceInMinutes, differenceInHours } from 'date-fns'
 
 interface DateValidation {
@@ -59,11 +59,21 @@ export class RuxTrack {
         this.initializeRows()
     }
 
+    @Listen('ruxtimeregionchange')
+    handleTimeRegionChange(e: CustomEvent) {
+        this.initializeRows()
+        e.stopPropagation()
+    }
+
     connectedCallback() {
         this._handleSlotChange = this._handleSlotChange.bind(this)
     }
 
-    calculateGridColumnFromTime(time: any) {
+    /**
+     * Tracks are displayed as a (CSS) grid of cells.
+     * Each cell can represent a minute or hour depending on the interval.
+     */
+    private _calculateGridColumnFromTime(time: any) {
         if (this.start) {
             const timelineStart = new Date(this.start)
 
@@ -141,7 +151,11 @@ export class RuxTrack {
         }
     }
 
-    initializeRows() {
+    /**
+     * Time Regions are dumb and don't know anything about the grid.
+     * The Track is responsible for lining up the Time Regions with the grid.
+     */
+    private initializeRows() {
         const children = [...this.el.children].filter(
             (el) => el.tagName.toLowerCase() === 'rux-time-region'
         ) as HTMLRuxTimeRegionElement[]
@@ -154,9 +168,9 @@ export class RuxTrack {
                 el.timezone = this.timezone
                 el.style.gridRow = '1'
                 el.style.visibility = 'inherit'
-                const gridColumn = `${this.calculateGridColumnFromTime(
+                const gridColumn = `${this._calculateGridColumnFromTime(
                     el.start
-                )} / ${this.calculateGridColumnFromTime(el.end)}`
+                )} / ${this._calculateGridColumnFromTime(el.end)}`
                 el.style.gridColumn = gridColumn
             } else {
                 if (!isHidden) {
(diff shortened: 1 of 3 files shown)


micahjones13 (approved):
I think the event is fine - in my mind it's less burdensome than a mutation observer anyway. More overhead would probably be needed to ensure it's done optimally (https://stackoverflow.com/questions/31659567/performance-of-mutationobserver-to-detect-nodes-in-entire-dom)

Paterma (approved):

kiley-mitti (approved):

packages/web-components/src/components/rux-timeline/rux-track/rux-track.tsx
```diff
@@ -63,7 +63,7 @@ export class RuxTrack {
         this._handleSlotChange = this._handleSlotChange.bind(this)
     }
 
-    calculateGridColumnFromTime(time: any) {
+    private _calculateGridColumnFromTime(time: any) {
```
jonathantneal, line 66: Would it be possible to document these methods? I wouldn’t necessarily know what makes them private or public at a glance.

packages/web-components/src/components/rux-timeline/rux-track/rux-track.tsx
```diff
@@ -154,16 +154,20 @@ export class RuxTrack {
...
                 el.style.gridColumn = gridColumn
             } else {
                 if (!isHidden) {
                     el.style.visibility = 'hidden'
                     console.error(isValid.error)
                 }
             }
+
+            el.addEventListener('ruxtimeregionchange', () => {
+                this.initializeRows()
```
jonathantneal, line 169: Will this reattach the handler every time there is a region change? If so, it might start slowing down as the component is used more.

FWIW, I do like the choice of events in this. My understanding is that browsers have made non-bubbling events very performant, plus they are still synchronous, and for added niceness the listeners can throw without stopping other handlers or the functions that dispatch them.

jonathantneal (changes requested):
Thanks for making this, @markacianfrani!

I have a concern about the event listeners, but the curiosity about the private-ness of methods is non-blocking.

markacianfrani:
> I think the event is fine - in my mind it's less burdensome than a mutation observer anyway. More overhead would probably be needed to ensure it's done optimally (https://stackoverflow.com/questions/31659567/performance-of-mutationobserver-to-detect-nodes-in-entire-dom)

YTM for looking into that!

packages/web-components/src/components/rux-timeline/rux-track/rux-track.tsx
markacianfrani: Good catch. They were in fact reattaching. 

My quick solution: attach the listener on the track host instead of each region element using Stencils @Listen sugar. 

alternative: I could set the ruxtimeregionchange to not bubble and go the same approach I had originally, but instead either 
1. Remove the event listener and add it every time (probably not performant?)
2. Manage the state of whether a listener has already been attached internally.

packages/web-components/src/components/rux-timeline/rux-track/rux-track.tsx
markacianfrani: Absolutely. Lmk what questions you still have. By default, every method is private. You'll know if something is public if you see Stencils Method decorator and documentation. We have very few public methods. 

Also, a nit on Stencil is that all public methods need to be async.

jonathantneal: Thank you!

packages/web-components/src/components/rux-timeline/rux-track/rux-track.tsx
```diff
@@ -171,24 +185,24 @@ export class RuxTrack {
...
+    //                     }}
+    //                     class={{
+    //                         cell: true,
+    //                         marker: i % 60 === 0,
+    //                     }}
+    //                 ></div>
+    //             ))}
+    //         </div>
+    //     )
+    // }
```
jonathantneal, line 205: Would you leave a `/** @TODO` note if you want to keep this commented-out code? Non-blocking.

jonathantneal (approved):
I left a request, but it is non-blocking.

Issue #625, opened 2022-08-09
Pass title attribute from button down to rux-icon when specified with rux-button


QGerrit:
I have a `rux-button` which, due to space requirements, only includes an icon. To give the user a description of what each button does, I set the html `title` attribute on each button, which works if I hover over the part of the `rux-button` which is not displaying the icon, but doesn't work when hovering over the icon itself.

```html
    <rux-button
      icon="palette"
      icon-only
      title="Title text"
    ></rux-button>
```

![image](https://user-images.githubusercontent.com/97202638/183756713-634da3f5-2f05-4489-a601-415a601b0494.png)

![image](https://user-images.githubusercontent.com/97202638/183756851-e352000d-f996-40b4-9176-ad6b88017476.png)

micahjones13:
Hey @QGerrit,

From the context, it seems like you're trying to add a tooltip to the button and icon, correct? If that's the case, then I would recommend staying away from the title attribute. See some reasons [here](https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute). 
Something you've probably noticed is that `rux-icon` actually assigns the title attribute - hence the "palette" showing up. This was an oversight on our part and will be removed in the future, so apologies for the confusion it has caused. 

What I would recommend to create a tooltip is to follow a CSS pattern such as this: https://www.w3.org/WAI/WCAG22/Techniques/css/C7. Also, if you would like a `rux-tooltip` or something similar in the future, our [GitHub discussions](https://github.com/RocketCommunicationsInc/astro/discussions/categories/ideas) are open for any ideas you may have!

Thanks, and please let me know if you have any other questions or need support.

jduncanRadBlue:
The same thing is happening to me.  I have an icon inside a button and when I hover over the icon, the tooltip is the icon's name and not the title which I have specified for the button.  How can I disable the icon's tooltip?

QGerrit:
That's fair, a `rux-tooltip` would be very useful! Having written them by hand before, it's really annoying to do so.

micahjones13:
The `title` attribute on the rux-icon has been removed in the above PR, and will be released with 7.0 since it's considered a breaking change. 
Right now, you can actually change the title attribute of the rux-icon if you use a slot on the button and set the label attribute. Here's an [example sandbox](https://codesandbox.io/s/fervent-bash-g3ycnp?file=/index.html). 
[...]

PR #627, opened 2022-08-10
Remove icon title

micahjones13:
## Brief Description

Removes the global title attribute from our rux icon. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-4317

## Related Issue
https://github.com/RocketCommunicationsInc/astro/issues/625

## Motivation and Context

Folks using the attribute title on button are having that title overriden when using an icon inside that button, due to us setting the title in the icon component. 

Title is not a recommended attribute in any standard, and should be avoided anyway. This takes care of that. 
Reference: https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-icon/rux-icon.tsx
+++ b/packages/web-components/src/components/rux-icon/rux-icon.tsx
@@ -45,12 +45,7 @@ export class RuxIcon {
 
         return (
             <Host>
-                <SVG
-                    class="icon"
-                    part="icon"
-                    size={this.size}
-                    title={this.iconLabel}
-                ></SVG>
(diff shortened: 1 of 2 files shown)


markacianfrani (changes requested):
This is a breaking change so it needs to go into next and include a changeset. This is also very much a grey area. I think this is good ammunition for more frequent major releases. 

On other hand, this gives us more time to prioritize rux-tooltip so that we could potentially have that ready in the migration strategy

PR #629, opened 2022-08-11
Update README browser chart with checkmarks

terryroe:
## Brief Description

Change the browser chart to use checkmarks (&check;) instead of X's to indicate a supported browser.

## Motivation and Context

The 'X' marks in the supported browser chart might indicate that the browsers were not supported.  A checkmark seems like a more obvious way to indicate support.

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


micahjones13 (approved):
Agreed, good catch. Thanks!

PR #644, opened 2022-08-16
Typography Utility Classes

markacianfrani:
## Brief Description

Publishes our new typography utility classes from our design tokens to the global css so peeps can use it.
* .rux-body-1
* .rux-body-1-bold
* .rux-body-2
* .rux-body-2-bold
* .rux-body-3
* .rux-body-3-bold
* .rux-control-body-1
* .rux-control-body-1-bold
* .rux-heading-1
* .rux-heading-1-bold
* .rux-heading-2
* .rux-heading-3
* .rux-heading-4
* .rux-heading-5
* .rux-heading-6
* .rux-display-1
* .rux-display-2
* .rux-monospace-1

## JIRA Link

ASTRO-4339

## Motivation and Context

Using raw design tokens for typography sucks

## Issues and Limitations

* Additional tickets will be created to clean up the import process a bit (ie not importing from /dist)
* These are imported by default without any way to opt out and will add to the bundle size. Because theres so few, it should be negligible, but providing a way to opt in/out is a high priority for me. Unfortunately, this has turned out to be very complicated.
* Documentation coming in separate ticket

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/lazy-pumas-trade.md
+++ b/.changeset/lazy-pumas-trade.md
@@ -1,0 +1,9 @@
+---
+"angular-workspace": minor
+"@astrouxds/angular": minor
+"astro-website": minor
+"@astrouxds/react": minor
+"@astrouxds/astro-web-components": minor
+---
+
+Adds typography utility classes
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -66,7 +66,7 @@ jobs:
         uses: cypress-io/github-action@v4
         with:
           working-directory: ./packages/web-components
-          install: ${{ (true && 'steps.cache-node-modules.outputs.cache-hit != 'true') || 'steps.cache-node-modules.outputs.cache-hit == 'true' }}
+          install: true
           start: npm run start.stencil
           wait-on: "http://localhost:3333"
           wait-on-timeout: "110"
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -69,4 +69,4 @@ jobs:
           install: true
           start: npm run start.stencil
           wait-on: "http://localhost:3333"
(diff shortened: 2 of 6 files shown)


kiley-mitti:
I love this! Non-blocking, possibly newbie question. Is it possible/desireable to use the design tokens in the [classes you've created](https://github.com/RocketCommunicationsInc/astro-design-tokens/blob/main/dist/css/classes/typography.css)? So that they're more of a compilation of the tokens rather than a reflection of them?

kiley-mitti (approved):

markacianfrani:
> I love this! Non-blocking, possibly newbie question. Is it possible/desireable to use the design tokens in the [classes you've created](https://github.com/RocketCommunicationsInc/astro-design-tokens/blob/main/dist/css/classes/typography.css)? So that they're more of a compilation of the tokens rather than a reflection of them?

Absolutely we can. Very trivially too. I'm thinking of revisiting that when we get to theming. You should be able to override a font size for example and have your classes work. 

My reasoning for not using tokens right now was that if you wanted to use this outside our components, it's nice to be able to just import the classes without having to import the tokens. I don't think that'll be a very common use case though

kiley-mitti:
> Absolutely we can. Very trivially too. I'm thinking of revisiting that when we get to theming. You should be able to override a font size for example and have your classes work.
> 
> My reasoning for not using tokens right now was that if you wanted to use this outside our components, it's nice to be able to just import the classes without having to import the tokens. I don't think that'll be a very common use case though

Awesome! I'm also not sure whether it'd be a useful thing or not.. A good thing to revisit later on. :) Thank you!

micahjones13 (approved):

Paterma (approved):

PR #645, opened 2022-08-17
Km 3866 switch toggle spacing

kiley-mitti:
## Brief Description

Add design tokens to rux-switch to bring in line with Figma design

**Question**: This is one of the components that requires actual height and width declarations. For those I used the --spacing tokens and some calculation but I'm wondering if that's the best path. Is there a better choice?

**Comment**: If you're wondering why components.d.ts was touched.. My stencil compiler keeps changing it when compiling. I had to bring it back inline with next.

## JIRA Link

[ASTRO-3866](https://rocketcom.atlassian.net/browse/ASTRO-3866)

## General Notes
I added inset shadows rather than borders to the thumb switch and rail in order to allow easier sizing via tokens.

## Motivation and Context

Part of the astro spacing adjustment initiative!

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components.d.ts
+++ b/packages/web-components/src/components.d.ts
@@ -271,6 +271,10 @@ export namespace Components {
          */
         "icon": string;
         /**
+           * The icon SVG's title attribute. Used for accessibility. If none is provided, the icon name will be used.
+         */
+        "label"?: string;
+        /**
           * The size of the icon. Can be 'extra-small', 'small', 'normal', 'large', 'auto' or any custom value ('30px', '1rem', '3.321em')
          */
         "size": | 'extra-small'
@@ -20582,6 +20586,10 @@ declare namespace LocalJSX {
          */
         "icon": string;
         /**
+          * The icon SVG's title attribute. Used for accessibility. If none is provided, the icon name will be used.
+         */
+         "label"?: string;
+         /**
           * The size of the icon. Can be 'extra-small', 'small', 'normal', 'large', 'auto' or any custom value ('30px', '1rem', '3.321em')
          */
         "size"?: | 'extra-small'
--- a/packages/web-components/src/components.d.ts
+++ b/packages/web-components/src/components.d.ts
@@ -271,10 +271,6 @@ export namespace Components {
          */
         "icon": string;
         /**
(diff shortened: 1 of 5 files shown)


packages/web-components/src/components.d.ts
```diff
@@ -271,7 +271,7 @@ export namespace Components {
          */
         "icon": string;
         /**
-          * The icon SVG's title attribute. Used for accessibility. If none is provided, the icon name will be used.
+           * The icon SVG's title attribute. Used for accessibility. If none is provided, the icon name will be used.
```
micahjones13, line 274: Are you running a `npm run build.stencil` at all? That command will generate this file. I can't remember if I've asked you that already, so sorry in advance.

kiley-mitti: Not that I know of! When I clone the repo and branch it the file is already there. :) https://github.com/RocketCommunicationsInc/astro/tree/next/packages/web-components/src
It seems that running start.stencil modifies it.

packages/web-components/src/components/rux-switch/rux-switch.scss
```diff
@@ -1,12 +1,11 @@
 @use '../../common/functional-components/FormFieldMessage/form-field-message.scss';
 :host {
     display: block;
-    padding-left: 1%;
 }
 .rux-switch {
     position: relative;
     display: flex;
-    height: 22px;
+    height: var(--line-height-sm);
```
markacianfrani, line 8: can we keep this hardcoded at 22px for now since its really a size token? if design changes `line-height-sm` (and they wont any time soon), it might unintended side effects. a more realistic scenario would be we change our line height tokens to be unitless values, which they should be but we're constrained at the moment.

markacianfrani (changes requested):

packages/web-components/src/components/rux-switch/rux-switch.scss
kiley-mitti: Sure! So it's actually 20px so to bring it inline with my other height 20px calculations I did a `calc(var(--spacing-4) + var(--spacing-1))` I figure if I'm consistent it'll be easier to find in future. Is that ok?

markacianfrani (approved):

PR #674, opened 2022-08-25
Bk 4320 update tree component

brikiernan:
Draft pr to review onConnectedCallback usage and initial style changes.

### DIFF
--- a/packages/web-components/src/components/rux-checkbox/rux-checkbox.scss
+++ b/packages/web-components/src/components/rux-checkbox/rux-checkbox.scss
@@ -1,11 +1,5 @@
 :host {
-    *,
-    *::before,
-    *::after {
-        box-sizing: inherit;
-    }
-
-    display: inline-block;
+    display: block;
 }
 
 .hidden,
@@ -13,38 +7,19 @@
     display: none;
 }
 
-.rux-form-field {
+:host::part(container) {
     display: flex;
-    flex-direction: column;
-    font-family: var(--font-body-1-font-family);
-    font-size: var(--font-body-1-font-size);
-    letter-spacing: var(--font-body-1-letter-spacing);
-    font-weight: var(--font-body-1-font-weight);
-    color: var(--color-text-primary);
+    min-height: 24px;
+    align-items: center;
+    gap: 8px;
+}
+
+:host([disabled]) {
+    opacity: var(--disabled-opacity);
 }
 
 .rux-checkbox {
-    display: flex;
-    position: relative;
-
-    &--indeterminate {
-        input[type='checkbox'] + label::after {
-            position: absolute;
-            display: flex;
-            content: '';
-            top: 5px;
-            width: 10px;
-            height: 5px;
-            transform: rotate(0deg);
-            border-right: 0;
-            border-top: 0;
-            border-bottom: 2px solid var(--color-background-interactive-default);
-            left: 4px;
-        }
-    }
-
-    input[type='checkbox'] {
-        -webkit-appearance: none;
+    &__input {
         appearance: none;
         margin: 0;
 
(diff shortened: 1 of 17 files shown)


packages/web-components/src/components/rux-tree-node/rux-tree-node.tsx
```diff
@@ -275,10 +285,16 @@ export class RuxTreeNode {
                 >
                     <div class="parent" tabindex="0">
                         {this._hasChildren && (
-                            <i
+                            <rux-icon
+                                class="icon"
```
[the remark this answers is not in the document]
brikiernan: I changed the class and updated the arrow closed/expanded logic

markacianfrani:
looking really good. can we:

1. NIT - rename nowrap custom prop to --white-space? so its more intuitive as a developer that the options are wrap, nowrap, etc. we'll just need to add some documentation in the tree story around that. 
2. NIT - rename the left/right slots to prefix/suffix to be consistent with what input calls them 

3. There's a little bit of jank happening when you expand/collapse a node

https://user-images.githubusercontent.com/1091342/187769063-eae669c5-3cb0-435e-afb6-29481cfd74b5.mov

markacianfrani:
4. We can actually remove the code that swaps the light theme status icons on selected. Previously, we needed to do that bc the selected bg color didnt have enough contrast with the status symbol icons but the new color doesnt have that issue. If you notice right now, when a node with a status icon is selected, the icons change to have a darker border.

brikiernan:
The four changes have been made as detailed.

packages/web-components/src/components/rux-tree-node/rux-tree-node.tsx
```diff
@@ -253,34 +256,56 @@ export class RuxTreeNode {
...
                                 class="arrow"
-                            ></i>
+                                onClick={this._handleArrowClick.bind(this)}
+                                size="1.25rem"
+                                icon={iconName}
+                            />
+                        )}
+                        {hasPrefix && (
+                            <span class="prefix">
+                                <slot name="prefix"></slot>
```
markacianfrani, line 298: since the slot is being conditionally rendered, we can't add slotted content after the component renders. 

For example, we should be able to do: 

```
<body>
        <rux-tree-node style="width: 300px;">whatup</rux-tree-node>
        <button id="add">add a prefix</button>
        <script>
            add.addEventListener('click', () => {
                const el = document.querySelector('rux-tree-node')
                const icon = document.createElement('rux-status')
                icon.setAttribute('status', 'critical')
                icon.setAttribute('slot', 'prefix')
                el.appendChild(icon)
            })
        </script>
    </body>
```

brikiernan:
@markacianfrani I updated the suffix and prefix slots to not conditionally render.

I'm not sure how that conflicts with the checkbox files got it here or how to resolve them.

packages/web-components/src/components/rux-tree-node/rux-tree-node.tsx
```diff
@@ -67,7 +83,7 @@ export class RuxTreeNode {
         this.setSelected(newValue)
     }
 
-    @Listen('keydown', { passive: true })
+    @Listen('keydown', { passive: false })
```
markacianfrani, line 86: do you know what the issue with this was?

brikiernan: Yes, if you have any scroll, and press the right or down key arrow, you do what that key should do in the component and scroll at the same time.

Basically if it's set to true the e.preventDefault does not work and you'll get an error saying something like "preventDefault won't work on passive events"

kiley-mitti:
Added the new tree parts for 7.0 here with Brian's excellent help! :)

packages/web-components/src/components/rux-checkbox/rux-checkbox.scss
```diff
@@ -46,6 +46,7 @@
     input[type='checkbox'] {
```
markacianfrani, line 46: can you remove the checkbox PR stuff from this PR please?

brikiernan:
@markacianfrani I removed the "wrap" prop and adjusted the css. Also, should I remove all the code I added in the test html file for rux-tree on this is ready to go?

packages/web-components/src/components/rux-tree-node/rux-tree-node.scss
```diff
@@ -1,67 +1,34 @@
...
-    user-select: none;
     display: block;
 }
 
 .tree-node {
     color: var(--color-text-primary);
     font-family: var(--font-body-1-font-family);
-    font-size: var(--font-body-1-font-size);
-    font-weight: var(--font-body-1-font-weight);
+    font-size: var(--font-size-base);
```
markacianfrani, line 8: any reason you're not using `--font-control-body-1-font-size`? and `--font-control-body-1-line-height`

brikiernan: I think I thought that was the correct token to use now, I'll change the size token back and I'll add the line-height on that class also

PR #676, opened 2022-08-25
Km 3842 rux textarea spacing alignment

kiley-mitti:
## Brief Description

Add spacing tokens and align to Astro design specifications. Additionally, in service to fixing an issue caused by padding in Firefox, I added a wrapper around textarea and added padding to that. Explained more thoroughly below.

## JIRA Link

[ASTRO-3842](https://rocketcom.atlassian.net/browse/ASTRO-3842)

## Motivation and Context

Part of the Astro spacing initiative!

## Issues and Limitations

Adding spacing to text areas adds a range of issues that you wouldn't expect, but it boils down to two issues:

**Issue 1: Padding**
Adding too much padding to a textarea breaks resizing in safari. The user becomes unable to get to the resizing handle for whatever reason.

After an extensive amount of testing, troubleshooting, and research I styled a new resizable wrapper around textarea. Safari doesn't have trouble with resize and padding when it is not attached to textarea. This works across all evergreen browsers but the scrollbar positioning is a little odd when a scrollbar exists, so it's an imperfect solution.

**Issue 2: Styling the resize handle**

Only ::-webkit browsers can target the resize handle, which means that it cannot be restyled in Firefox. I checked with design and they are ok with allowing the handle to be styled by the browser. I left my suggested solution for a styled handle commented out in the css, but this should be removed before merging.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/old-bottles-poke.md
+++ b/.changeset/old-bottles-poke.md
@@ -1,0 +1,9 @@
+---
+"angular-workspace": patch
+"@astrouxds/angular": patch
+"astro-website": patch
+"@astrouxds/react": patch
+"@astrouxds/astro-web-components": patch
+---
+
+rux-textarea - add spacing tokens and adjust code to fix firefox issue with resize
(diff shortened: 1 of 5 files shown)


markacianfrani (approved):
great find and work around w/ safari!

Issue #675, opened 2022-08-25
Rux-Dialog cannot keyboard to the deny button


tallen30:
We have a requirement that all actions can be done from the keyboard.

TabIndex needs to change, if possible link escape button to cancel too. 

https://github.com/RocketCommunicationsInc/astro/blob/main/packages/web-components/src/components/rux-dialog/rux-dialog.tsx
```
                                            <rux-button
                                                secondary={
                                                    confirmText.length > 0
                                                }
                                                onClick={_handleDialogChoice}
                                                data-value="false"
                                                hidden={!denyText}
                                                tabindex="-1"
                                                exportparts="container:deny-button"
                                            >
```

micahjones13:
Hey there,

This feature has just been merged and will be included in our next release this Thursday, 9/1. This feature includes:
- Hitting the Esc key will act as a "cancel" button press. 
- Tab Indexes have been updated on the buttons.

Please note that if you are using slots for the footer, this functionality will need to be manually set up by you - we are however looking at a possible solution to this in the future.

Thanks for bringing this up!

PR #677, opened 2022-08-25
feat(dialog): add better keyboard support

micahjones13:
## Brief Description

Adds tabindex's to the default (prop version) of the dialog footer. 
Adds the ability to trigger the deny button on an escape key press.
Added a section to dialog's SB page about keyboard interactions.

Added a commented out test - cypress does not allow for a keyboard press if it's not a form element. We will be switching these to playwright soon anyway.
Dialogs tests were still using modals - I changed that to be dialogs.

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-4456

## Related Issue
https://github.com/RocketCommunicationsInc/astro/issues/675
## General Notes

I've thought up a way to provide this keyboard functionality to folks using slots for the footer too, but it will need some more testing and is beyond the scope of this ticket. 
I've also added these changes to rux-modal

## Motivation and Context

GitHub issue - provides better keyboard accessibility to dialog. 

## Issues and Limitations

This does not add keyboard interactions by default to slot versions of dialog - that is on the external dev to add as of right now. A version with that functionality is in the works but is not ready yet.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-dialog/rux-dialog.tsx
+++ b/packages/web-components/src/components/rux-dialog/rux-dialog.tsx
@@ -81,10 +81,27 @@ export class RuxDialog {
     // confirm dialog if Enter key is pressed
     @Listen('keydown', { target: 'window' })
     handleKeyDown(ev: KeyboardEvent) {
-        if (ev.key === 'Enter') {
-            const button = this._getDefaultButton()
-            if (button) {
-                button.click()
+        if (this.open) {
+            const btns: NodeListOf<HTMLRuxButtonElement> = this.element.shadowRoot!.querySelectorAll(
+                'rux-button'
+            )
+            if (ev.key === 'Enter') {
+                //If enter is hit but the cancel/deny button is focused, we want to click that instead.
+                let activeEl: any = this.element.shadowRoot?.activeElement
+                if (activeEl && activeEl === btns[0]) {
+                    this._userInput = false
+                    btns[0].click()
+                } else {
+                    const button = this._getDefaultButton()
+                    if (button) {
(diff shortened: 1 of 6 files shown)


packages/web-components/src/components/rux-dialog/rux-dialog.tsx
```diff
@@ -235,16 +252,17 @@ export class RuxDialog {
                                                 onClick={_handleDialogChoice}
                                                 data-value="false"
                                                 hidden={!denyText}
-                                                tabindex="-1"
+                                                tabindex="0"
```
micahjones13, line 255: Did I get these tab indexes right? I'm not super familiar with them.

packages/web-components/src/components/rux-dialog/rux-dialog.tsx
```diff
@@ -81,10 +81,29 @@ export class RuxDialog {
     // confirm dialog if Enter key is pressed
     @Listen('keydown', { target: 'window' })
     handleKeyDown(ev: KeyboardEvent) {
-        if (ev.key === 'Enter') {
-            const button = this._getDefaultButton()
-            if (button) {
-                button.click()
+        // prevent this from running if the slots version is being used
+        if (this.open && !this.hasFooter) {
```
micahjones13, line 85: If the footer is a slot, this 'enter' press listener won't do anything since we don't grab what's being used in that slot. I figured it better to prevent this from running at all in that case.

markacianfrani (approved):

PR #680, opened 2022-08-29
refactor(web-components): use new token structure

markacianfrani:
## Brief Description

Updates our web components to use the new design token output structure

## General Notes

* removes the root-variables sass imports and uses [meta.module-variables](https://sass-lang.com/documentation/modules/meta#module-variables) instead. will revisit the approach to use sass here in a separate ticket. 
* removes the individual token files that just imported from node_mods 
* shaves off 2 KB from astro-web-components.css and 2KB from global-status-bar

## Motivation and Context

Our design token package has been updated to have a more intuitive file structure. Tokens are broken down by tier (ref/sys/comp).

Our tokens were previously exporting a custom "internal" directory that we consumed. All that did was change the selector from :root to `@mixin root-variables` so that we could use it as part of the GSB light theme workaround. That internal directory has been removed and now we use the same tokens as everyone else.

## Issues and Limitations
* No idea why I couldnt just import the index.css in global.scss like `@use '../../node_modules/@astrouxds/design-tokens/dist/css/index.css'`

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/package-lock.json
+++ b/packages/web-components/package-lock.json
@@ -16,7 +16,7 @@
             },
             "devDependencies": {
                 "@astrouxds/astro-figma-export": "~1.0.1",
-                "@astrouxds/design-tokens": "2.0.0-beta.12",
+                "@astrouxds/design-tokens": "2.0.0-beta.15",
                 "@astrouxds/storybook-addon-docs-stencil": "~1.0.10",
                 "@babel/core": "~7.13.16",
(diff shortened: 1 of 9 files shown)


Paterma (approved):

kiley-mitti (approved):
Makes sense to me!

FMorrison87 (approved):
Sounds like some savings!

PR #683, opened 2022-08-30
Km 3656 push button add parts

kiley-mitti:
## Brief Description

Added parts to rux-push-button to allow styling that was lost when the old css custom styles were deprecated. Changes part 'label' to 'container' for consistency with standard rux-button.

## JIRA Link

[ASTRO-3656](https://rocketcom.atlassian.net/browse/ASTRO-3656)

## General Notes
I came up with two ways to style rux-push-button when it is checked.

1) Add a part that only appears on <label> when checked='true' so that it can be targeted and styled
2) Use rux-push-button[checked]::part(label) to style the button when checked.

~~I decided to go with #1 since I believe it simplifies things for a developer using the component.~~
~~To that end I ultimately added 4 parts.~~
~~buttonOn, buttonOff, iconOn, iconOff~~

~~These cover the checked and unchecked button states for components that may be included in the button.~~

After checking out Mark's comments on rux-tabs I refactored this to use method 2.

The container part is not, strictly speaking, necessary as it occupies the same space as 'label'.. but I thought that it would be easier to understand at a glance.

I very much welcome thoughts on whether or not this is a good way to go about this.

## Motivation and Context

Returns style-ability to rux-push-button active states.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [x] Breaking change

## Checklist

- [x] This PR ~~adds or removes~~ changes a Storybook story.

### DIFF
--- a/.changeset/tidy-cycles-sleep.md
+++ b/.changeset/tidy-cycles-sleep.md
@@ -1,0 +1,9 @@
+---
+"angular-workspace": patch
+"@astrouxds/angular": patch
+"astro-website": patch
+"@astrouxds/react": patch
+"@astrouxds/astro-web-components": patch
+---
+
+rux-push-button add parts to allow styling which replaces deprecated css properties
--- a/.changeset/tidy-cycles-sleep.md
+++ b/.changeset/tidy-cycles-sleep.md
@@ -1,9 +1,0 @@
----
-"angular-workspace": patch
-"@astrouxds/angular": patch
-"astro-website": patch
-"@astrouxds/react": patch
-"@astrouxds/astro-web-components": patch
----
-
-rux-push-button add parts to allow styling which replaces deprecated css properties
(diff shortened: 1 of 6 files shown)


kiley-mitti:
After chatting with Mark I've replaced 'label as a part with 'container' so that styling rux-button and styling rux-push-button work the same way. This is a breaking change so I've also updated the changeset.

.changeset/tidy-cycles-sleep.md
```diff
@@ -0,0 +1,9 @@
+---
+"angular-workspace": patch
+"@astrouxds/angular": patch
+"astro-website": patch
+"@astrouxds/react": patch
+"@astrouxds/astro-web-components": patch
+---
+
+rux-push-button add parts to allow styling which replaces deprecated css properties
```
markacianfrani, line 9: think we can remove this changeset now?

markacianfrani (approved):
looks good! small comment on the 2nd changeset

.changeset/tidy-cycles-sleep.md
kiley-mitti: yup!

PR #693, opened 2022-09-01
Km 2080 add parts to input

kiley-mitti:
## Brief Description

Add part: input-field to rux-input to replace css custom styles that were deprecated 

## JIRA Link

[ASTRO-3655](https://rocketcom.atlassian.net/browse/ASTRO-3655)

## General Notes

Added a part to the element that wraps <input> since that is where most of the styling takes place. In order to style the wrapper on input:focus, a dev must use :focus-within but that pseudo class is well supported and it is clear from the parts descriptions that 'input' and 'input-field' style different parts so it is probably ok. 

## Motivation and Context

Some CSS custom properties were deprecated. This adds the ability to style the input element back to rux-input

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

- [x] This PR ~~adds or removes~~ changes a Storybook story.

### DIFF
--- a/packages/web-components/src/components/rux-input/rux-input.tsx
+++ b/packages/web-components/src/components/rux-input/rux-input.tsx
@@ -24,6 +24,7 @@ let id = 0
  * @part help-text - The help text element
  * @part icon - The icon displayed when toggle-password prop is set
  * @part input - The input element
+ * @part input-field - the styled wrapper around the input element
  * @part label - The input label when `label` prop is set
  * @part required - The asterisk when required is true
  * @part prefix - The container of the prefix slot
@@ -297,6 +298,7 @@ export class RuxInput implements FormFieldInterface {
                     ) : null}
 
                     <div
+                        part="input-field"
                         class={{
                             'rux-input': true,
                             'rux-input--focused': this.hasFocus,
--- a/.changeset/many-cycles-cheat.md
+++ b/.changeset/many-cycles-cheat.md
@@ -1,0 +1,9 @@
+---
+"angular-workspace": patch
+"@astrouxds/angular": patch
+"astro-website": patch
+"@astrouxds/react": patch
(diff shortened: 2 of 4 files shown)


packages/web-components/src/components/rux-input/rux-input.tsx
```diff
@@ -297,6 +298,7 @@ export class RuxInput implements FormFieldInterface {
                     ) : null}
 
                     <div
+                        part="input-field"
```
micahjones13, line 301: I almost feel like `input-wrapper` would be a better name? I care like a 3/10 though so this is non-blocking. @markacianfrani , whatdya think

micahjones13 (approved):

packages/web-components/src/components/rux-input/rux-input.tsx
kiley-mitti: I was oscillating between wrapper and field. I would also normally gravitate toward wrapper but I went into some of our other components for inspiration and noticed that others were using -field to denote something wrapping a form item. Of course in those cases, the field usually includes the label. I was aiming at consistency but am happy to change it! :)

kiley-mitti:
going to close this since it was supposed to go into main. :) stay tuned!

packages/web-components/src/components/rux-input/rux-input.tsx
markacianfrani: I'd go for consistency in this case. This one is difficult. I think 'wrapper' as a term is semantic junk food, but I still use it all the time. The difference between a part and a class though is we need to think about support. We can change the internals (add more wrappers inside wrappers), but we always need to keep this part so it needs to have a more semantic meaning. Having said that . . . I dont have any better semantic names in mind for this particular one

PR #695, opened 2022-09-02
Bk 4225 4492 fix label bugs

brikiernan:
## Brief Description

Removed the hasLabelSlot State and refactored the conditional rendering of the label "prop" vs the label "slot". 

Recommending the removal of the label "prop" in v7. Added the "Deprecated" tag in the prop comments. 

Removed dead classes.
 
## JIRA Link

[Ticket 4225](https://rocketcom.atlassian.net/browse/ASTRO-4225)
[Ticket 4492](https://rocketcom.atlassian.net/browse/ASTRO-4492)

## Motivation and Context

This change eliminates the multiple renders when the checkbox is checked or unchecked causing the hasLabelSlot boolean state to switch back to false when it should be true. This is the root cause of the 2 known bug tickets. 

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/common/functional-components/FormFieldMessage/FormFieldMessage.tsx
+++ b/packages/web-components/src/common/functional-components/FormFieldMessage/FormFieldMessage.tsx
@@ -35,29 +35,6 @@ const FormFieldMessage: FunctionalComponent<FormFieldMessageProps> = (
     }
 
     return null
-
-    // function shouldShowHelpText() {
-    //     return helpText && !errorText
-    // }
-
-    // function shouldShowErrorText() {
-    //     return errorText ? true : false
-    // }
-
-    // return (
-    //     (shouldShowHelpText() ? (
-    //         <div class="rux-help-text" part="help-text">
-    //             {children}
-    //             {helpText}
-    //         </div>
-    //     ) : null) ||
-    //     (shouldShowErrorText() ? (
-    //         <div class="rux-error-text" part="error-text">
-    //             {children}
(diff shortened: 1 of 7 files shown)


micahjones13 (approved):

markacianfrani (changes requested):
When there's no label prop or slot, we don't want the label and its right margin to render. 

* If the label prop is set and there is no slotted content, it should display the label as text with the correct margin.
* If the label prop is set and there is slotted content, it should display the slotted content as the label with the correct margin.
* If the label prop is not set and there is no slotted content, it should not display any label and there should not be any right margin.
* If the label prop is not set and there is slotted content, it should display the slotted content as the label with the correct margin.

The issue relates to the `renderHiddenInput` method that's used to add native `<form>` support. A hidden input with the checkbox value is added, but it's being considered as slotted content. Checkbox is also unique in that it also removes its hidden input when unchecked, causing the `hasSlot` utility to incorrectly return true because it detects the hidden input as slotted content.

markacianfrani (approved):
form field refactor looks good! just a comment about a comment. 

after this gets merged, can you also open a PR against next with these changes? I imagine there will be a decent amount of conflicts bc of the spacing changes that are in 7.0

brikiernan:
> after this gets merged, can you also open a PR against next with these changes? I imagine there will be a decent amount of conflicts bc of the spacing changes that are in 7.0

Yes! I was going to ask about that.

PR #701, opened 2022-09-06
fix: adding in specific heights to rux-tabs to fix a bug on setting t…

FMorrison87:
# Height Fix on rux-tabs

## Brief Description
fix: Updated styling on rux-tabs to make the height specific to the 'small' and 'large' component props.

## JIRA Link

[ASTRO-4228](https://rocketcom.atlassian.net/browse/ASTRO-4228)

## Motivation and Context

This fixes a bug that @brikiernan discovered where tabs will expand to the height of a container if the parent div has a set height. The Figma docs indicate that small tabs should have a fixed height of 60px and large a fixed height of 80px. These heights were not set in the stylesheet so I added them in via spacing tokens. Additionally, I needed to add in box-sizing: border-box on the .rux-tabs class in order for the padding and the new set height to play nicely with each other.

My only concern is a question of intent regarding where the tabs will sit now in the scenario in which Brian identified the bug, a parent div with a fixed height. With these new changes the tabs will sit in the top-left, but I'm not sure if design would like something different or if the issue is irrelevant as the developer can style the tabs to any part of the parent, technically, since that is outside the scope of the component itself.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


markacianfrani (approved):
just nix that packages/web-components/playwright-report/index.html file and g2g!

markacianfrani:
@FMorrison87 also could you please add a patch changeset?

PR #703, opened 2022-09-07
feat: Accordion Component

FMorrison87:
## Brief Description
The Accordion component is a brand new web component based off of the new design in Figma. The accordion component consists of a parent component `rux-accordion` and a child component `rux-accordion-item` that is nested within. The purpose of the `rux-accordion` is to have a block of title content that shows or hides content below the block on click. These items are stacked on top of each other in a list format. The title section can take two distinct pieces of information via slots, a left-most slot intended for an icon or indicator with a named slot called 'prefix', and a slot to the right of this named 'label' intended to take title text. The content section is controlled via the default slot. These slots can technically contain any valid html, but the intention from design is for the optional prefix slot to contain an Astro `rux-icon` component, and for the label slot to contain title text. The default slot can contain any valid html including Astro web components which will be stacked by default.

The developer can use parts to further style the `rux-accordion` and `rux-accordion-item` components if desired.

The developer can add props to `rux-accordion` and `rux-accordion-item` to control the behavior of the component. These props are:
**rux-accordion**
- disallow-multiple: only allows one child item to be open at a time
- expanded: expands all child items by default
- disabled: disables all child items by default

**rux-accordion-item**
- expanded: expands item by default
- disabled: disables item by default

## JIRA Link

[JIRA 2136](https://rocketcom.atlassian.net/browse/ASTRO-2136)

## Motivation and Context
The accordion component was completely redesigned in Figma and as such it needed to be built from the ground-up as a new web component. The process of designing this web component involved drafting an API prior to any code being written.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


markacianfrani (approved):

brikiernan (approved):
Great work @FMorrison87  and @kiley-mitti !!

PR #713, opened 2022-09-12
Km 3658 rux tree 6 add parts

kiley-mitti:
## Brief Description

Add parts to rux-tree-node to replace deprecated css custom properties as a means of styling 

## JIRA Link

[ASTRO-3658](https://rocketcom.atlassian.net/browse/ASTRO-3658)

## Motivation and Context

some css custom properties were deprecated, leading to a new way of thinking about how one would style the web components. To that end, these shadow parts are meant to return styling. ability to developers

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

- [] This PR adds or removes a Storybook story.

### DIFF
--- a/packages/web-components/src/components/rux-tree-node/rux-tree-node.tsx
+++ b/packages/web-components/src/components/rux-tree-node/rux-tree-node.tsx
@@ -23,6 +23,8 @@ let id = 0
 /**
  * @slot (default) - The parent node content
  * @slot node - Renders a child node within the current node
+ * @part item - The individual tree node
+ * @part icon - The open/closed indicator arrow that exists on parent nodes
  */
 export class RuxTreeNode {
     private componentId = `node-${++id}`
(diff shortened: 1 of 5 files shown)


packages/web-components/src/components/rux-tree-node/rux-tree-node.tsx
```diff
@@ -23,6 +23,8 @@ let id = 0
 /**
  * @slot (default) - The parent node content
  * @slot node - Renders a child node within the current node
+ * @part item - The individual tree node
```
micahjones13, line 26: I personally like 'node' here more than item. I don't see an issue with overlapping the naming with the slot. Since it's called 'rux-tree-node', just seems more natural that the part you'd be styling is the 'node'. 

Again, I don't want to block this with something like that lol. @markacianfrani whaddya think

markacianfrani: ya I agree I think `node` is better of the two

kiley-mitti: Happy to change it :) we'll have to change it from item to node in Brian's branch for next as well. I'll ping him about it since I don't know where he's at on that one.

micahjones13 (approved):

markacianfrani (approved):

PR #730, opened 2022-09-16
Light theme updates

markacianfrani:
## Brief Description

implements recent 6.0 light theme changes 

* tree node's arrow fill color on hover
* tree node's border left when selected
* fixes regression in segmented button background color on hover
* push button's text color 
* push button's icon color 
* adds borders around notification 
* fixes regression in clock border color
* swaps red-500 with 700 for errors 

## JIRA Link

ASTRO-4594

## Issues and Limitations

6.0 is already using a subset of the new 7.0 naming convention. Some of these changes require new tokens, but instead of publishing a whole separate, short-lived release of our 6.0-flavored tokens, I just created the ones that we needed locally in the vars.scss. 

Anything that uses an svg as a background image (FormFieldMessage error icon and checkbox) doesn't theme switch bc we can't use css custom props when an svg is inlined, as far as I could figure out. I have a hacky work around for tree node, but I'll address the others in a separate PR. 

New VRTs will need to be generated 

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/unlucky-foxes-pretend.md
+++ b/.changeset/unlucky-foxes-pretend.md
@@ -1,0 +1,8 @@
+---
+"@astrouxds/astro-web-components": minor
+"@astrouxds/angular": minor
+"astro-website": minor
+"@astrouxds/react": minor
+---
+
+Updated light theme styles
--- a/packages/web-components/.storybook/preview-head.html
+++ b/packages/web-components/.storybook/preview-head.html
@@ -21,6 +21,10 @@
         background: var(--color-background-base-default);
     }
 
+    .sbdocs .docs-story {
+        background: var(--color-background-surface-default);
+    }
+
     .sbdocs.sbdocs-h1 {
         font-size: 2.125rem;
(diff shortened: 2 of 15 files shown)


micahjones13 (approved):
For the form field message inlined SVG, could we not replace the `currentColor` fill with a custom prop? Ie, 
```
                    <path
                        fill-rule="evenodd"
                        clip-rule="evenodd"
                        d="that long string"
                        fill="var(--color-critical-or-whatever)"
                    ></path>
```

markacianfrani:
> For the form field message inlined SVG, could we not replace the `currentColor` fill with a custom prop? Ie,
> 
> ```
>                     <path
>                         fill-rule="evenodd"
>                         clip-rule="evenodd"
>                         d="that long string"
>                         fill="var(--color-critical-or-whatever)"
>                     ></path>
> ```

If its an actual SVG, ya, but not if its a bg image as far as I know

PR #767, opened 2022-09-26
fix(dialog): event detail now correct after reopening same modal

micahjones13:
## Brief Description

The `e.detail` emitted by `rux-dialog` would not reset to it's default value after the event was emitted. Because of this, if you opened the dialog, clicked confirm, it would emit true. If you opened that same dialog again and clicked off to close it, it would emit true again even though no choice was made. 

Added a test for this case.

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-4635

## Motivation and Context

Fixes an inconsistency in the event detail of dialog.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/metal-fishes-share.md
+++ b/.changeset/metal-fishes-share.md
@@ -1,0 +1,9 @@
+---
+"angular-workspace": patch
+"@astrouxds/angular": patch
+"astro-website": patch
+"@astrouxds/react": patch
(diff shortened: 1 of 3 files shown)


markacianfrani (approved):
just needs a changeset

PR #800, opened 2022-09-30
fix(astro 7.0): fixing non-local styles on parent overriding component styles

FMorrison87:
## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/package-lock.json
+++ b/packages/web-components/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@astrouxds/astro-web-components",
-    "version": "7.0.0-rc.2",
+    "version": "7.0.0",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "@astrouxds/astro-web-components",
-            "version": "7.0.0-rc.2",
+            "version": "7.0.0",
             "license": "MIT",
             "dependencies": {
                 "@floating-ui/dom": "~1.0.1",
--- a/packages/web-components/src/components/rux-accordion/rux-accordion-item/rux-accordion-item.scss
+++ b/packages/web-components/src/components/rux-accordion/rux-accordion-item/rux-accordion-item.scss
@@ -17,6 +17,7 @@ summary {
     box-sizing: border-box;
     display: flex;
     flex-direction: row;
+    text-align: left; /*needed to fix problem where parent will override text-align*/
     align-items: center;
     border-radius: var(--radius-base);
     background-color: var(--color-background-surface-header);


micahjones13 (approved):
Good start - I think chaos testing is def the way to go from here, but I still can't think of a world where we can stop any and all un-intended inheritance. Is there a confluence doc created yet that documents all of this? 
PS. that PR description was written very, very well. Props to you!

kiley-mitti (approved):
approving a thing I worked on is weird, but looks good to me! Really good description of the problem and current fix. :)

packages/web-components/src/components/rux-card/rux-card.scss
```diff
@@ -30,11 +30,19 @@
...
     line-height: var(--font-heading-3-line-height);
     box-sizing: border-box;
+    text-align: left; // Solves the issue of a parent element with a different text-align overriding.
 }
 
 .rux-card__body {
     padding: var(--body-padding);
     box-sizing: border-box;
+    font-size: var(
+        --font-size-base
```
markacianfrani, line 40: lets swap the --base tokens with the typography tokens for body-1:

```
    font-family: var(--font-body-1-font-family);
    font-size: var(--font-body-1-font-size);
    font-weight: var(--font-body-1-font-weight);
    line-height: var(--font-body-1-line-height);
    letter-spacing: var(--font-body-1-letter-spacing);
```

packages/web-components/src/components/rux-pop-up/rux-pop-up.scss
```diff
@@ -1,6 +1,9 @@
 :host {
     display: block;
     width: min-content;
+    font-size: var(--font-size-base); //prevents icon shift from font-size declaration on parent.
```
markacianfrani, line 4: can we move this to the internal `rux-popup` class? I just tried to quick and it seemed to work, were there any specific issues with this one?

packages/web-components/src/components/rux-card/rux-card.scss
markacianfrani: actually this applies to just about anything that uses --font-size-base as well

packages/web-components/src/components/rux-status/rux-status.scss
```diff
@@ -13,4 +13,10 @@
 }
 .status-wrapper {
     width: var(--spacing-3);
+    font-size: var(
```
markacianfrani, line 16: this is causing a pretty significant regression in the monitoring icon and monitoring progress icons. the status symbol is pushed much farther down. easiest fix might just be to override this in those components, ie 
```
.rux-advanced-status__status rux-status::part(status) { 
  line-height: 0; 
}
```

FMorrison87: Styles added.

packages/web-components/src/components/rux-pop-up/rux-pop-up.scss
FMorrison87: Yup, works great!

packages/web-components/src/components/rux-card/rux-card.scss
FMorrison87: This has been applied to all components where applicable, but I kept line-height as `--line-height-base` as this caused weird icon shift in `rux-global-status-bar` when I tried to apply the line-height for font-body-1.

markacianfrani (approved):

PR #811, opened 2022-10-06
opacity token correction

kiley-mitti:
## Brief Description

Changes --disabled-opacity token to --opacity-disabled in rux-checkbox
I also ran a general search for --disabled-opacity to ensure that there weren't other instances. I didn't find any

## JIRA Link

[ASTRO-4686](https://rocketcom.atlassian.net/browse/ASTRO-4686)

## Motivation and Context

I just noticed the token was wrong and thus it wasn't giving checkbox standard opacity for disabled elements.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/polite-pianos-allow.md
+++ b/.changeset/polite-pianos-allow.md
@@ -1,0 +1,5 @@
+---
+"@astrouxds/astro-web-components": patch
+---
+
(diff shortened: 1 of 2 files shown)


markacianfrani (approved):
good catch! I'm concerned about how this got in and past our VRTs but we'll push this out soon

Issue #821, opened 2022-10-11
Pop up menu doesn't close after selection


run2xs:
The default behavior on the popup menu should be that it automatically closes after a selection has been made.  If a client wants to enable mult-select on the menu or for some other reason wants to control it manually, then a property should be used to indicate that behavior.   In that case, the popup won't close until it is programatically done so.

We don't need anything rolled back for this.  Our workaround is fine for now, but I think this should be the default behavior.

micahjones13:
Because changing the default behavior to pop-up closing on a selection would be considered a breaking change, we've added a `close-on-select` prop to `rux-pop-up` that will provide this functionality. We'll definitely evaluate making this the default behavior in the next major version, though.
This new prop will be available in the release tomorrow, version 7.0.3.
Thanks for bringing this up!

PR #828, opened 2022-10-11
feat(popup): added single-select prop

micahjones13:
## Brief Description

Adds a single-select prop to rux-pop-up that allows the pop up to close when ruxmenuselected is emitted. 
Adds tests for this new prop

Bouns: 
- The playwright issue with `.toHaveAttribute` has been fixed, so I refactored all the tests to use this instead of the workaround that we were using.

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-4733

## Related Issue
https://github.com/RocketCommunicationsInc/astro/issues/821
## Motivation and Context

GitHub issue. 

## Issues and Limitations

We should make this the default behavior in the future, and instead add a `multi-select` prop or something similar. 
Because making pop up close after a selection would now be a breaking change, we can do this for now.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components.d.ts
```diff
@@ -12133,6 +12133,10 @@ export namespace Components {
           * Opens the pop up and returns true.
          */
         "show": () => Promise<true>;
+        /**
+          * When provided, will close the pop-up when a single selection is made.
+         */
+        "singleSelect": boolean;
```
markacianfrani, line 12139: what about close-on-select instead? looks good to me though! will this fire a close event too?

micahjones13: Yup, the `open` prop watch is handling the open/close events so this will emit a close event too. I also renamed it to be close-on-select

kiley-mitti (approved):
Nice, that's a great feature. :)

Issue #839, opened 2022-10-13
Popup menu sometimes doesn't close when close-on-select property is set


run2xs:
@micahjones13 Thanks for adding the close-on-select property to the rux-pop-up.  The popup menu will close correctly if my menu item action navigates to a different route.  This did not work before, so it's an improvement.  However, I have another menu item that just changes the light/dark theme by changing the class on the body element.  When I choose this one, the menu item does not close.

markacianfrani:
do you mind pasting just your markup (or some minimal version of it) so we can get a better idea of your use case?

run2xs:
Here's an edited version of the HTML:

`<rux-global-status-bar include-icon="false" appName="{{xx}}" appVersion="{{xxx}}">
  <rux-pop-up #ruxPopUp close-on-select id='xxx-popup-menu' slot='left-side'>
    <rux-icon slot='trigger' size="small" icon="apps" aria-controls="xxx-popup-menu"></rux-icon>
    <rux-menu>
      <rux-menu-item *ngFor='let navPage of navPages' (click)='navigateTo(navPage.route)'>{{navPage.label}}</rux-menu-item>
      <rux-menu-item-divider></rux-menu-item-divider>
      <rux-menu-item *ngIf="isDarkTheme" (click)='changeTheme()'>Set Light Theme</rux-menu-item>
      <rux-menu-item *ngIf="!isDarkTheme" (click)='changeTheme()'>Set Dark Theme</rux-menu-item>
      <rux-menu-item-divider *ngIf="(isAuthenticated | async) && isAuthEnabled"></rux-menu-item-divider>
      <rux-menu-item (click)='logout()'>Logout</rux-menu-item>
    </rux-menu>
  </rux-pop-up>
  <atlas-ui-system-mode></atlas-ui-system-mode>
  <rux-clock timezone='UTC'></rux-clock>
</rux-global-status-bar>`

Here's the call that doesn't close the popup:

` changeTheme() {
    this.isDarkTheme = !this.isDarkTheme;
    const themeColorClass = this.isDarkTheme ? 'dark-theme' : 'light-theme';
    document.body.className = themeColorClass;
    //todo: remove this when Astro fixes the close bug
    this.ruxPopUp['el'].hide();
  }
`

run2xs:
Not sure why I'm losing all the formatting above.  I'm attaching a file with the same content.

[...]

PR #841, opened 2022-10-13
feat(menu): allow for nested content in menu-items

micahjones13:
## Brief Description

Updates rux-menu to handle nested content inside of rux-menu-items. 
Previously, if you had content that wasn't a rux-menu-item or nested content inside a menu-item, the events wouldn't fire like they should. 

## Related Issue
https://github.com/RocketCommunicationsInc/astro/issues/839

## Motivation and Context

Allows for rux-menu to be used as a container, but also provides selection functionality when used with rux-menu-items. 

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-menu/rux-menu.tsx
+++ b/packages/web-components/src/components/rux-menu/rux-menu.tsx
@@ -11,26 +11,28 @@ export class RuxMenu {
     /**
      * Emits when a rux-menu-item is selected. Emits the rux-menu-item selected in the event detail.
      */
-    @Event({ eventName: 'ruxmenuselected' })
+    @Event({ eventName: 'ruxmenuselected', bubbles: true })
     ruxMenuSelected!: EventEmitter
 
     connectedCallback() {
         this.el.addEventListener('click', (e) => {
-            this._handleSelected(e.target as HTMLRuxMenuItemElement)
+            this._handleSelected(e.target as HTMLElement)
         })
     }
 
-    private _handleSelected(item: HTMLRuxMenuItemElement) {
+    private _handleSelected(item: HTMLElement) {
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/rux-menu/rux-menu.tsx
```diff
@@ -11,26 +11,28 @@ export class RuxMenu {
...
-            if (el === item && !el.disabled) {
-                item.selected = true
-                this.ruxMenuSelected.emit(item)
-            }
-        })
+        if (item.hasAttribute('disabled')) return
+        //prevent items that are nested in a disabled rux-menu-item from emitting
+        if (item.closest('rux-menu-item')?.hasAttribute('disabled')) return
+        const menuItems = Array.from(this.el.children)
+        if (menuItems.length > 1) {
```
micahjones13, line 29: This `> 1` check is for instances where menu is being used as a container. Like with the div and two buttons example you gave earlier @markacianfrani . Without this, it will change the single `rux-menu-item` to a selected state when interacted with which was weird looking when it was being used as a container/menu.

packages/web-components/src/components/rux-menu/rux-menu.tsx
```diff
@@ -11,26 +11,28 @@ export class RuxMenu {
...
+            menuItems.forEach((el) => {
+                el.removeAttribute('selected')
+            })
+            item.closest('rux-menu-item')?.setAttribute('selected', '')
+        }
+        this.ruxMenuSelected.emit(item)
```
micahjones13, line 35: This still selects the precise clicked item. So if we have a nested span, it'll emit that span in the `e.detail`

PR #854, opened 2022-10-24
Km 4780 rux notification message display

kiley-mitti:
## Brief Description

Removed the message prop from rux-notification. This is a major breaking change.

## JIRA Link

[ASTRO-4780](https://rocketcom.atlassian.net/browse/ASTRO-4780)

## General Notes

**The Problem**
When a named slot is added to <rux-notification> the fallback prop 'message' for the default slot is no longer displayed.

For example:
` <rux-notification open status="standby" message="Insert better example here"> `
`   <rux-icon icon="apps" slot="prefix"></rux-icon> `
` </rux-notification> `

Should display the 'apps' icon to the left of a standby icon and the message 'Insert better example here' to the right.
However, it doesn't display the message.

**The Reason**
This is actually expected web component behavior, even though at a glance it doesn't seem like that. the prop "message" displays as a fallback for the default slot. When default slot is empty it should display. However, the default slot is not actually empty. There are hidden spaces and carriage returns surrounding <rux-icon slot="apps"></rux-icon> that are filling the default slot. This can be tested by rewriting the code above on one line with no spaces like so:
` <rux-notification open status="standby" message="Insert better example here"><rux-icon icon="apps" slot="prefix"></rux-icon></rux-notification> `

The 'message' should then appear as expected.

[Source](https://stackoverflow.com/questions/53182353/fallback-content-on-an-unnamed-slot-is-never-displayed)

**The Solution**
One possibility would be to write a method that seeks out and ignores these spaces and carriage returns.. but that could get complicated and messy quite quickly.

Instead, I've chosen to remove the message prop and change the instructions for this component. I believe that this is a better choice on a couple of levels.

1) I think that adding the message inside of the <rux-notification>Notification Message here!</rux-notification> is more intuitive in the long run. Using props as fallback for text can get confusing in corner cases. For example, what if a dev utilizes both the message prop and the default slot?
` <rux-notification message="This is my notification">J/K It's here</rux-notification> `
What is the expected outcome? Should it render both? Just the default slot? Just the message?

2) Our [storybook on notification](https://astro-components.netlify.app/?path=/docs/components-notification--default-story#default-slot) actually already calls out that things put into the default slot will override 'message' in this case it's just doing it for an unexpected reason.

3) Other comparable web-component libraries that use slots for this component handle the main content of an alert or notification component in a similar manner. [shoelace](https://shoelace.style/components/alert),[ kor](https://kor-ui.com/components/notifications), or do not list a default slot: [Carbon](https://web-components.carbondesignsystem.com/?path=/docs/components-notifications--inline)

## Issues and Limitations

While this is a simple fix, it is going to be a big ugly breaking change when it goes into production, so if we decide to implement it later and just add some documentation for now, or go a different direction entirely, I can test out other solutions. :)

## Types of changes

- [x] Bug fix
- [ ] New feature
- [x] Breaking change

## Checklist

### DIFF
--- a/.changeset/spicy-buses-play.md
+++ b/.changeset/spicy-buses-play.md
@@ -1,0 +1,9 @@
+---
+"@astrouxds/astro-web-components": major
+---
+
+WHAT: rux-notification: remove message prop
+
+WHY: when other named slots are entered into <rux-notification></rux-notification> the message prop is overwritten by unintentional blank spaces and carriage returns. A better pattern might be to have devs enter their message into the default slot.
+
+HOW TO MIGRATE: If you were previously using message="my notification" to add your notification message, please move the message to the default slot within rux-notification such as <rux-notification>My Message</rux-notification>
--- a/packages/web-components/src/components.d.ts
+++ b/packages/web-components/src/components.d.ts
@@ -12080,10 +12080,6 @@ export namespace Components {
          */
         "hideClose": boolean;
         /**
-          * Message for the notification banner.
-         */
-        "message": string;
-        /**
           * Set to true to display the Banner and begin countdown to close (if a close-after Number value is provided).
          */
         "open": boolean;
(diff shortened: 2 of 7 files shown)


markacianfrani:
I agree we should just remove the prop. Keeping this open for now bc I want to think about some ways to improve our process of deprecating and breaking changes.

markacianfrani:
Created ASTRO-4935 to circle back on this after we decide if we're going to change our versioning strat. Closing for now

PR #855, opened 2022-10-25
Mc fm 4558 remove jest unit tests

markacianfrani:
## Brief Description

* Refactors some tests to use fixtures 
* Removes some unnecessary or duplicate tests. 

## Motivation and Context

This PR got a little too ambitious and Im not sure if we should just scrap it. 

It adds an AstroPage fixture, which can do all of the setup for us via a `load` method. 

A test like:
```
   test('it renders', async ({ page }) => {
        await setBodyContent(
            page,
            `
        <rux-button>Hello</rux-button>
    `
        )
        const el = page.locator('rux-button').first()
        await expect(el).toBeVisible()
        await expect(el).toHaveClass('hydrated')
    })
```

becomes 

```
    test('it renders', async ({ astroPage }) => {
        const el = await astroPage.load('<rux-button>foo</rux-button>')
        await expect(el).toHaveAttribute('hydrated')
    })
```

To make a little more easier to read what's going on, I've moved the template markup to its own const: 
```
    test('it renders', async ({ page }) => {
       const template = `<rux-button>foo</rux-button>`
       const el = await astroPage.load(template)
        await expect(el).toHaveAttribute('hydrated')
    })
```

The one thing I kept seeing repeated is that we need to use a locator after setting content. this fixture will return the first child you put in it so that you can immediately use it. 

This is totally opt in.

## Reasons Not To Merge
* too much abstraction
* half finished/broken windows. I wasn't able to entirely replacement every test to this new fixture. I underestimated how much time that would take.

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


micahjones13 (approved):
The AstroComponentPage and AstroFormPage are much needed abstractions imo. I really like this!

PR #864, opened 2022-10-28
fix(pop-up): Popup auto placement no longer set by default, enabled via prop

FMorrison87:
## Brief Description

This fix changes some core functionality in pop-up that was causing the pop-up to migrate around its trigger depending on scroll position and allowed space between the trigger and the browser window. This behavior can now be set via prop and default to false.

## JIRA Link

[ASTRO-4779](https://rocketcom.atlassian.net/browse/ASTRO-4779)

## General Notes

This functionality is handled mostly using [floating-ui](https://floating-ui.com/). In order to make this toggle-able a prop was added that the developer can add onto the component to initiate the behavior (basically turning it back into what it was before), but without this prop the pop-up will open based on settings using the `placement` prop. Some tweaks were made to the logic within the floating-ui functions and helper functions to make this possible. Thanks to @kiley-mitti for her help in working through some of the logic!

## Motivation and Context

After conferring with design, it was learned that this behavior where the pop-up moves depending on scroll or position to the browser edge was not necessarily desired, but since it's in there already instead of removing it entirely it should default to OFF but then be toggle-able. 

## Types of changes

- [x] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


markacianfrani (commented):
can we keep the default behavior to on? otherwise this will be a breaking change and have to go into 8.0. might need to rename the prop to something though since we can't have bools default to true. maybe disableAutoUpdate?

FMorrison87:
> 

@markacianfrani that change is made. Default is now to have that behavior be on and the bool turns it off.

packages/web-components/src/components/rux-pop-up/rux-pop-up.tsx
```diff
@@ -55,6 +55,11 @@ export class RuxPopUp {
      */
     @Prop() placement: ExtendedPlacement = 'auto'
 
+    /**
+     * Turns autoUpdate on or off which makes the pop-up move to stay in view based on scroll. Defaults to false.
+     */
+    @Prop({ mutable: true, reflect: true }) disableAutoUpdate: boolean = false
```
markacianfrani, line 61: we don't need to mark this as mutable since it doesnt change. I also dont think we need to reflect it but I dont feel strongly. Reflecting is absolutely necessary if we're doing :host[attr] styles but thats not too relevant here .

FMorrison87: ok all set!

PR #876, opened 2022-10-31
chore: delete astrouxds from repo

markacianfrani:
## Brief Description

end of an era


micahjones13 (approved):
let there be less clutter

PR #903, opened 2022-11-10
Mj 4867 pop up overflow

micahjones13:
## Brief Description

The `strategy` prop on rux-pop-up was not being applied correctly. This now applies the correct `position: ` to the pop-up. 
This fixes an issue where a pop-up inside a parent that had `position: absolute; overflow: hidden` would hide the pop-up. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-4867

## Motivation and Context

Pop-up was being hidden when the above conditions where being met. 

## Issues and Limitations

This still does not fix an issue where a `transform` on a parent will still hide some or all of the pop-up. This is due to `transform` creating it's own stacking context, and thus the pop-up gets rendered behind it. 

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/flat-cups-sit.md
+++ b/.changeset/flat-cups-sit.md
@@ -1,0 +1,8 @@
+---
+"@astrouxds/astro-web-components": patch
+"angular-workspace": patch
+"@astrouxds/angular": patch
+"@astrouxds/react": patch
+---
+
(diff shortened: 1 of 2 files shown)


kiley-mitti (approved):
Love it! Solves the overflow:hidden on parent container issue I was having. :)

markacianfrani (approved):

PR #904, opened 2022-11-10
Mj 4868 dialog msg

micahjones13:
## Brief Description

Fixed an issue where the message prop on rux-dialog would not show if being used with slots for header or footer.
This was due to an issue where even though the message(default) slot was empty, it was taking in whitespace and therefore thinking it wasn't empty. This fix conditionally renders the slot based on if a default slot is being used.

This relates to @kiley-mitti 's PR of https://github.com/RocketCommunicationsInc/astro/pull/854. I think it's the exact same issue that the default message slot on notification is having. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-4868

## Related Issue
https://github.com/RocketCommunicationsInc/astro/pull/854
## Motivation and Context

Message prop can now be used on dialog at any point.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-dialog/rux-dialog.tsx
+++ b/packages/web-components/src/components/rux-dialog/rux-dialog.tsx
@@ -236,9 +236,15 @@ export class RuxDialog {
                                     }}
                                     part="message"
                                 >
-                                    <slot onSlotchange={this._handleSlotChange}>
-                                        {message}
-                                    </slot>
+                                    {this.hasMessage ? (
+                                        <slot
+                                            onSlotchange={
+                                                this._handleSlotChange
+                                            }
+                                        ></slot>
+                                    ) : (
+                                        <div>{message}</div>
+                                    )}
                                 </div>
                             </div>
                             <footer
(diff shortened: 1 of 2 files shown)


kiley-mitti (approved):
Cool Micah! I think this works really well since it doesn't have to conditionally render the slot. It should work exactly the same on message, let me go back and make a PR on that. I'm going to leave the other one up where I suggest we deprecate message as a prop, but I do think this is a great first step!

packages/web-components/src/components/rux-dialog/rux-dialog.tsx
```diff
@@ -186,9 +187,14 @@ export class RuxDialog {
     }
 
     private _handleSlotChange() {
+        console.log('handle slot change trigger')
```
markacianfrani, line 190: think these are failing the tests

markacianfrani (approved):
should we write a test for this (answer can be no)? also src/index.html still has your setup.

packages/web-components/src/components/rux-dialog/rux-dialog.tsx
micahjones13: yeah, I pushed this up just for Kiley to see. Will clean it up

PR #905, opened 2022-11-11
emulate changes micah made to rux-dialog

kiley-mitti:
## Brief Description

This is a followup to the rux-notification issue [#854](https://github.com/RocketCommunicationsInc/astro/pull/854) while I still believe that the ultimate solution is to remove the message prop, @micahjones13 has come up with a brilliant interim solution that is not a breaking change and does fix the issue. I've basically taken his work on rux-dialog and emulated it here. :) 

## JIRA Link

[ASTRO-4780](https://rocketcom.atlassian.net/browse/ASTRO-4780)

## Related Issue
[#904](https://github.com/RocketCommunicationsInc/astro/pull/904)

## General Notes

Thanks Micah!

## Motivation and Context

This will allow the message prop to render if the default slot isn't filled. EVEN if there is another slot inside the component that is being used.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/heavy-kings-smell.md
+++ b/.changeset/heavy-kings-smell.md
@@ -1,0 +1,5 @@
+---
+"@astrouxds/astro-web-components": patch
+---
+
+rux-notification: fix but where message prop was not showing if another slot was slotted in the component
--- a/.changeset/heavy-kings-smell.md
+++ b/.changeset/heavy-kings-smell.md
@@ -2,4 +2,4 @@
 "@astrouxds/astro-web-components": patch
 ---
 
-rux-notification: fix but where message prop was not showing if another slot was slotted in the component
+rux-notification: fix a bug where message prop was not showing if another slot was slotted in the component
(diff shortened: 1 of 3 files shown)


.changeset/heavy-kings-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrouxds/astro-web-components": patch
+---
+
+rux-notification: fix but where message prop was not showing if another slot was slotted in the component
```
micahjones13, line 5: I think there's an extraneous 'but' here

kiley-mitti: There's always an extraneous 'but' with me around.
Fixed!

micahjones13 (approved):

markacianfrani (approved):
can we do a quick audit of other components and make sure this isn't happening anywhere else?

kiley-mitti:
on it @markacianfrani I have a good idea of what we're looking for

Issue #907, opened 2022-11-11
rux-icon in Storybook shows a label prop


micahjones13:
**Describe the bug**
On https://astro-components.netlify.app/?path=/story/components-icons--default-story there is a 'label' control. rux-icon doesn't have a label prop. 

**To Reproduce**
Steps to reproduce the behavior: Go to https://astro-components.netlify.app/?path=/story/components-icons--default-story

**Current behavior**
There is a label control in SB on rux-icon.

**Expected behavior**
There shouldn't be a label control.

PR #908, opened 2022-11-11
(fix rux-icon): Removing incorrect label prop from rux-icon storybook

FMorrison87:
## Brief Description

The Storybook story for `rux-icon` was incorrectly showing that it had a label prop. This has been removed from its story.

## JIRA Link

[ASTRO-4872](https://rocketcom.atlassian.net/browse/ASTRO-4872)

## Related Issue
https://github.com/RocketCommunicationsInc/astro/issues/907
## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/clever-hounds-watch.md
+++ b/.changeset/clever-hounds-watch.md
@@ -1,0 +1,7 @@
+---
+"@astrouxds/angular": minor
+"@astrouxds/react": minor
+"@astrouxds/astro-web-components": minor
+---
+
+Fixing rux-icon in Storybook incorrectly displaying a label prop.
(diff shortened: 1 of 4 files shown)


.changeset/clever-hounds-watch.md
```diff
@@ -0,0 +1,7 @@
+---
```
micahjones13, line 1: I don't think this requires a changeset since it's just a docs change. @markacianfrani whaddya think

micahjones13: I forgot he's at a conference all day. Let's just nix this changeset (you can just delete the .md file) since it doesn't actually effect any styling or component functionality

FMorrison87: @Micah, ok all set!

PR #913, opened 2022-11-14
(fix rux-status): Status has no default

FMorrison87:
## Brief Description

If the user does not supply a 'status' to the component then nothing renders; it has no default state. This fix gives it a default state of 'normal' if nothing is passed in to status.

## JIRA Link

[ASTRO-4880](https://rocketcom.atlassian.net/browse/ASTRO-4880)

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-status/rux-status.tsx
+++ b/packages/web-components/src/components/rux-status/rux-status.tsx
@@ -39,7 +39,9 @@ export class RuxStatus {
     }
 
     connectedCallback() {
-        if (this.status) this.validateStatus(this.status)
+        this.status
+            ? this.validateStatus(this.status)
+            : (this.status = 'normal')
     }
 
     render() {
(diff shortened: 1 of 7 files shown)


packages/web-components/src/components/rux-status/rux-status.tsx
```diff
@@ -39,7 +39,9 @@ export class RuxStatus {
     }
 
     connectedCallback() {
-        if (this.status) this.validateStatus(this.status)
+        this.status
```
micahjones13, line 42: Because this is changing the status internally, we get a console warning from stencil: 
```
@Prop() "status" on <rux-status> is immutable but was modified from within the component.
```
Could we instead change the status prop to have a default of 'normal'? I'll make a comment at that location, too.

micahjones13: Welp since github won't let me comment on it since it's not a 'changed file', it's on line 19. If we do: 
```
@Prop({ reflect: true }) status?: Status = 'normal'
```
then the situation of having no status passed in is solved before render and we won't get that Stencil warning. 
Any downsides to this that I may have missed?

micahjones13 (changes requested):

micahjones13:
Could we give this a 'minor' changeset? A case could be made for patch, but I see this as more of a feature/enhancement than patch fix 🤷

PR #926, opened 2022-11-16
style(seg-btn): added pointer cursor

micahjones13:
## Brief Description

Adds `cursor: pointer` to an enabled rux-segmented-button.

I wanted to also add `pointer: not-allowed` to a disabled seg button, but because we're setting `pointer-events: none` on `:host[disabled]`, it doesn't work. We'd have add a wrapper or something and set that to `cursor: not-allowed`. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-4877

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/warm-rice-divide.md
+++ b/.changeset/warm-rice-divide.md
@@ -1,0 +1,8 @@
+---
+"@astrouxds/astro-web-components": patch
+"angular-workspace": patch
+"@astrouxds/angular": patch
+"@astrouxds/react": patch
+---
+
+Added a `cursor: pointer` to an enabled rux-segmented-button.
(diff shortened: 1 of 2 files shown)


kiley-mitti (approved):
👈 ☝️ 👉 👇

micahjones13:
> > I wanted to also add `pointer: not-allowed` to a disabled seg button, but because we're setting `pointer-events: none` on `:host[disabled]`, it doesn't work. We'd have add a wrapper or something and set that to `cursor: not-allowed`.
> 
> if we move pointer-events:none to something like:
> 
> host[disabled] .rux-segmented-button{pointer-events:none} and then have :host[disabled] {pointer: not-allowed}
> 
> Would that work?

I did try that, and it's as close as I got. The issue being that if you add a click event listener to rux-segmented-button it'll still fire that attached listener even when disabled. That being said, I'm not entirely sure that's a problem? It's not hard for a user to check if it's disabled before firing the click event, so maybe it's fine. I'll bring it up in our astro-dev channel and see what everyone thinks!

markacianfrani (approved):
changeset here as well

PR #937, opened 2022-11-21
added parts for app meta and the default slot

kiley-mitti:
## Brief Description

I've added new parts to aux-global-status-bar to open up more styling options for developers.

## JIRA Link

[ASTRO-4884](https://rocketcom.atlassian.net/browse/ASTRO-4884)

## Related Issue

[Discussion #993](https://github.com/RocketCommunicationsInc/astro/discussions/933)

## Motivation and Context

The GSB's default slot is difficult to center on the page. Since the GSB uses a flex-box model, if the right-slot is less wide than the left-slot + App container (as it most often is) then the content in the default slot will center on its surrounding container but NOT on the page. See [Example](https://dpv114.csb.app)

## Issues and Limitations

Without adding a bunch of markup, this is a difficult thing to fix within the component itself. Every section of the GSB is slottable and we have no way to really know what developers will choose to put in each slot.

To that end, I added two parts: one on the container that wraps the app-data section and another on the container that wraps the default slot.

With these two extra style-able section, it should be significantly less challenging to place the default slot content where one would want it.

I don't LOVE the name I've given to the default slot wrapper part and would be very open to suggestion.

There is a centering proof-of-concept on center.html in the GSB test folder. This should be removed before we merge.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/fuzzy-feet-protect.md
+++ b/.changeset/fuzzy-feet-protect.md
@@ -1,0 +1,5 @@
+---
+"@astrouxds/astro-web-components": patch
+---
+
+rux-global-status-bar add parts for app-meta and default slot styling
--- a/packages/web-components/src/components/rux-global-status-bar/appMeta/appMeta.tsx
+++ b/packages/web-components/src/components/rux-global-status-bar/appMeta/appMeta.tsx
@@ -10,7 +10,7 @@ export const AppMeta: FunctionalComponent<AppMetaProps> = (
     { domain, name, version },
     children
 ) => (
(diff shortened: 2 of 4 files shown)


markacianfrani (approved):
good stuff! I think this raises a larger issue with the component itself and we should reevaluate the value add of gsb as a whole. devs shouldnt have to do this, but as a short term solution, we should at least offer the ability to do so.

PR #942, opened 2022-11-21
Monorepo configuration updates

jonathantneal:
## Brief Description

- ~Updates the root `package.json#engines.npm` from `~8.5` to `^8.5`, allowing newer versions of `npm` v8.x.~ **Merged!**
- ~Updates the root `package.json#engines.node` from `^16` to `^16 || ^18`, allowing `node` v18.x.~. **Merged!**
- ~Updates the Web Components `package.json#devDependencies` to remove an unused dependency on `puppeteer`.~ **Merged!**
- ~Updates the Web Components `.prettierrc` configuration to match the rest of the project’s formatting of JSON and applies the effect.~ **Merged!**
- Updates any `package.json` dependencies to reflect the versions installed from `npm install` given their current semver range.
- Updates any `package-lock.json` files to reflect the updated state of `package.json` files.
- Updates the react `package.json#dependencies` to synchronize `@types/node`, which prevented the build from running.

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


packages/angular-workspace/projects/angular/package.json
```diff
@@ -13,6 +13,6 @@
   },
   "dependencies": {
     "@astrouxds/astro-web-components": "^7.3.0",
-    "tslib": "^2.3.0"
+    "tslib": "^2.4.1"
```
markacianfrani, line 16: If the default angular 12 freshie install is tslib@2.3.0, is this going to cause issues for someone when they go to pull in astrouxds/angular?

jonathantneal: No and yes.

The default installation of this package can install any version of `tslib`, currently ranging from v2.3.0 to v2.4.1. This can depend on their machine, existing caching, the semver of other packages they are using, and their package manager. The `^` range explicitly allows any of these versions to be acceptable. The tooling reconciles these things for a user.

markacianfrani (approved):

PR #949, opened 2022-11-22
test: updating event tests and cleanup, up to rux-datetime

micahjones13:
## Brief Description

Cleaned up tests and added some event testing. This is an ongoing effort, more PRs will be opened with more components. This spans accordion thru datetime. 

Deletes any 'it renders' tests, and in cases where that was the only test (like card and container), deletes the test file. 
## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-4895

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


markacianfrani (approved):
lil merge conflict with clock now

PR #958, opened 2022-11-29
Focus State Work

FMorrison87:
## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/index.html
```diff
@@ -19,5 +19,17 @@
         <link rel="stylesheet" href="/build/astro-web-components.css" />
     </head>
 
-    <body></body>
```
markacianfrani, line 22: cleanup

packages/web-components/src/components/rux-button/rux-button.scss
```diff
@@ -45,10 +45,24 @@
...
     }
 
-    &:focus {
-        outline: none;
+    &:focus-visible {
+        outline: var(--focus-outline);
+        outline-offset: var(--focus-offset);
     }
 
+    // &:focus-visible::after {
```
markacianfrani, line 53: cleanup

packages/web-components/src/components/rux-button/rux-button.scss
```diff
@@ -25,7 +25,7 @@
     line-height: var(--font-control-body-1-line-height);
     letter-spacing: var(--font-control-body-1-letter-spacing);
     white-space: nowrap;
-    overflow: hidden;
+    // overflow: hidden;
```
markacianfrani, line 28: cleanup

packages/web-components/src/global/_variables.scss
```diff
@@ -28,6 +28,14 @@
     --normal-border: #56f000;
     --off-fill: #a4abb6;
     --off-border: #a4abb6;
+
+    //focus variables
+    --focus-ring-color: var(--color-palette-pink-200);
```
markacianfrani, line 33: just need to update to use the latest tokens. may need to rebase main to pull the latest if they arent in here 

--color-border-focus-default - The color of a focused element’s outline
--border-width-focus-default The thickness of a focused element’s outline
--spacing-focus-default  - The amount of space between a focused element’s outline and edge

FMorrison87: All set!

packages/web-components/src/components/rux-radio-group/test/radio-group.spec.ts
```diff
@@ -191,28 +191,166 @@ test.describe('Radio-group-with-form', () => {
...
         `
         await page.setContent(template)
     })
     test('renders label prop and slot', async ({ page }) => {
         //Arrange
-        const ruxRadioProp = await page.locator('rux-radio-group').first()
+        const ruxRadioProp = page.locator('rux-radio-group').first()
         const labelProp = ruxRadioProp.locator('.rux-label')
-        const ruxRadioSlot = await page.locator('rux-radio-group').nth(1)
+        const ruxRadioSlot = page.locator('rux-radio-group').nth(1)
```
markacianfrani, line 223: this test is failing now. think it might have to do with the changes to the template content. rather than try and use the same template for every test, its okay to split it up so each test has its own template setup, even if it means duplicating some.

markacianfrani (approved):

markacianfrani (changes requested):
can we get changesets for every component? for each component, they can just be 'added focus state' but we should call out things like in radio group, added keyboard support as a separate changeset too

kiley-mitti:
I've gone through the list of changed files and added a changes for each component we added focus styles for, and also a separate one for each keyboard control change, and one for the global style change.  Double check me but I think it's all there. :)

markacianfrani (approved):

PR #963, opened 2022-12-02
Fm km 4878 tooltip mvp

markacianfrani:
## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


markacianfrani:
If you add a tooltip to some plain text like: 
```
Hello my name is <rux-tooltip message="yo">whatup</rux-tooltip>
```
the content gets cursor pointer, which makes me think I can click it. It also prevents me from being able to select/copy the content. Im thinking we should remove this. 

But there's a larger UX concern here...in this example, there should probably be some visual affordance that **whatup** is a tooltip that you can hover over. Icon maybe? or is this example bad juju? Going to raise this up with design to make sure we capture that in our docs.

packages/web-components/package.json
```diff
@@ -44,7 +44,8 @@
         "@floating-ui/dom": "~1.0.1",
         "@stencil/core": "~2.18.0",
         "date-fns": "~2.21.1",
-        "date-fns-tz": "~1.3.3"
+        "date-fns-tz": "~1.3.3",
+        "docker": "~1.0.0"
```
markacianfrani, line 48: not sure if this was intentional?

FMorrison87: D'oh. Fixed.

packages/web-components/src/components/rux-slider/test/index.html
```diff
@@ -104,7 +104,7 @@ <h2>Prop/Slot Testing</h2>
             <script>
                 const log = document.getElementById('log')
                 const ticks = document.getElementById('ticks')
-                ticks.axisLabels = ['0', '25', '50', '75', '100']
+                ticks.axisLabels = ['Low', 'Med', 'High', 'Oh Noes!']
```
markacianfrani, line 107: whats going on here?

kiley-mitti: @markacianfrani @FMorrison87 OH NO! I'm sorry. 🤣 Jonathan and I were having a side discussion about how ticks work and get labeled in slider and I wrote up something and shared screen but didn't think I committed it! How embarrassing. Let me fix.

PR #977, opened 2022-12-12
add vrt light theme

markacianfrani:
## Brief Description

adds light theme vrts
also noticed that monitoring icon has some flake now from the latest PR. tried to resolve that in this PR by adding some sleep and checking for Roboto, unsure if it totally fixes the issue.

## JIRA Link

ASTRO-4904

## Issues and Limitations

we could use fixtures to eliminate some of this duplication, but I decided to do the dumbest possible implementation here so that its clear to everyone what exactly is going on with these tests. we can always refactor later

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


jonathantneal (changes requested):
This looks great. There’s a lot going on, so I’m passing over all the image changes.

I also have a few recommendations:
- extract as much of the repetitive evaluation as possible into a fixture.
  - Reference: [Playwright Fixtures](https://playwright.dev/docs/test-fixtures#with-fixtures)
- check `whenDefined` for elements, rather than network idle?
  - Reference: [Stencil Methods ](https://stenciljs.com/docs/methods)

markacianfrani:
> This looks great. There’s a lot going on, so I’m passing over all the image changes.
> 
> 
> 
> I also have a few recommendations:
> 
> - extract as much of the repetitive evaluation as possible into a fixture?
> 
>   - Reference: [Playwright Fixtures](https://playwright.dev/docs/test-fixtures#with-fixtures)
> 
> - check `whenDefined` for elements, rather than network idle (plus the document font check)?
> 
>   - Reference: [Stencil Methods ](https://stenciljs.com/docs/methods)

1. Id like to avoid fixtures right now and add them in later. I had some fixtures before and ended up ripping them out because I abstracted the wrong things too prematurely. Once we have a solid idea of everything we need, I plan to refactor a bit. Ideally you'd just write your tests once for dark mode, then playwright would run them again in light. 
2. I'll rip out the roboto check. It didn't end up fixing the flake anyway. Will address that separately.

PR #994, opened 2022-12-22
add focus state for close button/keyboard control

kiley-mitti:
## Brief Description

Rux-notification could not be focused since it did not have a tab index. Added that and keyboard controls and then our focus outline state.

## JIRA Link

[ASTRO-5078](https://rocketcom.atlassian.net/browse/ASTRO-5078)

## Motivation and Context

Focus state general tweaking

## Issues and Limitations

We need to revisit some of our components that use overflow:hidden in any way. Sometimes this cuts off focus state which isn't great. In notification specifically we should revisit the idea that notifications should be truncated with no way to see the rest of the notification.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-notification/rux-notification.scss
+++ b/packages/web-components/src/components/rux-notification/rux-notification.scss
@@ -84,6 +84,13 @@
     margin-left: auto;
     display: flex;
 }
+.rux-notification-banner__close {
+    &:focus-visible {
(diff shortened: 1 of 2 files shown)


FMorrison87 (approved):
Looks great! Nice work Kiley.

PR #1008, opened 2023-01-11
fix(rux-checkbox): fix bug where the invisible <input> component was interfering with certain layouts

kiley-mitti:
## Brief Description

The <input> element in the Shadow DOM of rux-checkbox is opacity:0 and position: absolute. However, since no parent element was position:relative, the <input> was just kind of going wherever it felt like on the page and wreaking havoc. I added position:relative to the label element which contains both our custom input and the actual <input> in order to contain it.

## JIRA Link

[ASTRO-5148](https://rocketcom.atlassian.net/browse/ASTRO-5148)

## Motivation and Context

Minor bug fix to address some layout abnormalities when using many rue-checkboxes, as you might in a table of events.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/modern-items-refuse.md
+++ b/.changeset/modern-items-refuse.md
@@ -1,0 +1,5 @@
+---
+"@astrouxds/astro-web-components": patch
+---
+
+fix(rux-checkbox) add position relative to label to contain input element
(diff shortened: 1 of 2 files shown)


markacianfrani (approved):
good catch!

PR #1026, opened 2023-01-19
test fixture

markacianfrani:
## Brief Description

Lots of file changes because I had to rename our snapshot files and regenerate them. 

We want to have VRTs for light theme in addition to dark theme. Currently, our light theme tests require us to duplicate our dark theme tests: 

```
    test('has no visual regression @vrt @dark', async ({ page }) => {
        await page.goto(`/src/components/rux-accordion/test/basic/index.html`)
        await expect(page).toHaveScreenshot()
    })

    test('has no visual regression @vrt @light', async ({ page }) => {
        await page.goto(`/src/components/rux-accordion/test/basic/index.html`)
        await page.evaluate(() => {
            document.body.classList.add('light-theme')
        })
        await expect(page).toHaveScreenshot()
 }
```
this worked fine for simple basic tests, however, once we start adding more tests for things like hover and focus state, it will get out of hand quickly having to duplicate those as well. 

this PR creates a new fixture that will automatically run all the of VRTs twice, one for dark, one for light, off one single test and reduces the above to:

```
    test.use({ component: 'rux-accordion' })

    test('has no visual regression @vrt', async ({ astroVRTPage }) => {
        await expect(astroVRTPage).toHaveScreenshot()
    })
```

## General Notes

It works by using Playwright's [Parameters](https://playwright.dev/docs/test-parameterize#parameterized-projects). I've split our project into three: 1 for regular e2e, 1 for dark theme, 1 for light theme. The project config in playwright.config.ts is responsible for changing the `theme` parameter for each. The fixture is responsible for checking if the theme is light, and then appends the `light-theme` body class.

## Issues and Limitations

If we go beyond simple, basic test folder structure, we'll need to add another param for that. 

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components/rux-button/test/basic/button.vrt.spec.ts
```diff
@@ -1,16 +1,9 @@
...
+    test.use({ component: 'rux-button' })
 
-    test('has no visual regression @vrt @light', async ({ page }) => {
-        await page.goto(`/src/components/rux-button/test/basic/index.html`)
-        await page.evaluate(() => {
-            document.body.classList.add('light-theme')
-        })
-        await expect(page).toHaveScreenshot()
+    test('has no visual regression @vrt', async ({ astroVRTPage }) => {
+        await expect(astroVRTPage).toHaveScreenshot()
```
markacianfrani, line 7: is this too much of an abstraction? the actual goto step is moved into the fixture. one downside I can see to this is it might be confusing if youre unaware and trying to figure out where the test setup is. 

but the big value add here is not having to write duplicate tests for dark and light theme. if we move goto to a fixture, we can automatically append the light theme class after load.

markacianfrani: another con--having to add another var if/when we expand into more detailed test dirs other than 'basic'.

PR #1028, opened 2023-01-20
fix(rux-progress-monitoring-icon) restyle to match rux-monitoring-icon

kiley-mitti:
## Brief Description

This is a restyle of rux-progress-monitoring-icon web component to bring the styles inline with rux-monitoring-icon. The two were styles quite differently which made them look odd next to each other in the GSB. The [GRM Dashboard](https://www.astrouxds.com/grm-service-ux-design/grm-dashboard/) is a sample of where they might be seen together in the GSB.

## JIRA Link

[ASTRO-5105](https://rocketcom.atlassian.net/browse/ASTRO-5150)

## Motivation and Context

Oddly there is no figma example of rux-progress-monitoring-icon so its styles were never (as far as I know) created by design. This is a draft PR to show design what it would look like if it matched rux-monitoring-icon

## Issues and Limitations

This is a draft until it gets approved by design. :)

## Types of changes

- [x] Bug fix
- [ ] New feature
- [x] Breaking change - it does pretty significantly change the style of rux-progress-monitoring-icon although it should not change the overall footprint/size.

## Checklist

### DIFF
--- a/.changeset/gold-melons-prove.md
+++ b/.changeset/gold-melons-prove.md
@@ -1,0 +1,5 @@
+---
+"@astrouxds/astro-web-components": minor
+---
+
+fix(rux-progrress-monitoring-icon) restyle to match rux-monitoring-icon
--- a/packages/web-components/src/components/rux-monitoring-progress-icon/rux-monitoring-progress-icon.scss
+++ b/packages/web-components/src/components/rux-monitoring-progress-icon/rux-monitoring-progress-icon.scss
@@ -35,8 +35,6 @@
 
 rux-status {
     position: absolute;
-    top: -0.25rem;
-    left: -0.25rem;
     margin: 0;
 }
 
@@ -68,12 +66,12 @@ rux-status {
     font-size: var(--font-body-2-font-size);
     font-weight: var(--font-body-2-font-weight);
     letter-spacing: var(--font-body-2-letter-spacing);
+    line-height: var(--line-height-sm);
     text-align: center;
     text-overflow: ellipsis;
     white-space: nowrap;
-    line-height: 1.2;
     overflow: hidden;
-    margin-top: 1rem;
+    margin-top: 0;
     width: 100%;
     max-width: 6.25rem;
 }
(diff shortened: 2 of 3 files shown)


FMorrison87 (approved):
This is a great change!

markacianfrani:
this might need some more design thought, esp around how the badge floats. 

localhost_3333_ (2)
I like how the badge here floats, but it doesnt on the monitoring icon. they should be consistent. 
(also think the 100% is too close to the border. we didnt have a vrt covering 100%)

localhost_3333_ (3)
the problem with that though is that if you have a single icon in the right slot in gsb, the badge will get cut off.

kiley-mitti:
> this might need some more design thought, esp around how the badge floats.
> localhost_3333_ (2) I like how the badge here floats, but it doesnt on the monitoring icon. they should be consistent.

Oh right... I need to add the math for sizing to the progress monitoring icon. Frances added it to the monitoring icon some time ago to fix the badge cut off problem with GSB.

> (also think the 100% is too close to the border. we didnt have a vrt covering 100%)

Agree on this. Steve G and I actually had a conversation here and decided that for now, it's probably ok because we can't make the text smaller (design system reasons). When they get this into figma they'll play with it and perhaps find a slight tweak to solve.

> localhost_3333_ (3) the problem with that though is that if you have a single icon in the right slot in gsb, the badge will get cut off.

Yeah, we fixed this for monitoring icon, but not for progress monitoring icon. I'll go in and make that revision.

I'm wondering if progress monitoring icon is meant to have notifications at all. I don't know if it is meant to or if that's just a remnant of it being copied over from monitoring icon.

packages/web-components/src/components/rux-monitoring-progress-icon/rux-monitoring-progress-icon.tsx
```diff
@@ -136,6 +137,10 @@ export class RuxMonitoringProgressIcon {
         }
     }
 
+    componentDidRender() {
+        this.handleNotificatonWidth()
```
markacianfrani, line 141: one issue with this approach (and also in monitoring icon maybe), is that the width doesn't get smaller if the notifications get lower. it will correctly expand if you go from 99 -> 999, but if you go from 999 -> 99, it still has the maximum width. this means that when the notification count changes, it'll cause the layout to jank around as the widths adjust. I think that part is unavoidable unfortunately, but we should at least look into having it update when you go from 999 -> 99.

PR #1039, opened 2023-01-24
Km 5331 add container part to rux-container

kiley-mitti:
## Brief Description

Added a 'container' part to the outermost div of the rux-container. I also added scrollbar stylings.

## JIRA Link

[ASTRO-5331](https://rocketcom.atlassian.net/browse/ASTRO-5331)

## Motivation and Context

It was brought to our attention by @brikiernan that a very common use case for containers was to hold scrollable content. It was difficult to style the height of rux-container in such a way that would allow the inner parts to align properly and the body to scroll. Adding part='container' to the surrounding div should allow for more flexibility and easier access to change height properties/add Flexbox or grid stylings for layout.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/proud-wolves-divide.md
+++ b/.changeset/proud-wolves-divide.md
@@ -1,0 +1,5 @@
+---
+"@astrouxds/astro-web-components": patch
+---
+
+feat(rux-container) add container part to the encompassing div to allow for extra styling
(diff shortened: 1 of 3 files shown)


micahjones13 (approved):

markacianfrani (changes requested):
can you change the changeset to minor plz since we're adding a new part? the scroll styles would be patch but we can just roll it all in one. also, can you mark all the packages instead of just web-comps? if we only mark web comps, changesets won't bump the wrappers when we version.

kiley-mitti:
sure thing!

markacianfrani (approved):

PR #1053, opened 2023-01-30
fix(rux-pop-up): Placement of pop-up when trigger shifts

FMorrison87:
## Brief Description

An issue was brought to our attention where if the pop-up was open, and something in the application caused the trigger to move, either external to the trigger or by modifying the styling on the trigger itself, the placement of the pop-up would not follow this shift. This fix further utilizes [floating-ui's autoUpdate animationFrame option](https://floating-ui.com/docs/autoUpdate#animationframe) to watch the component and detect any shifts in trigger placement. This will appropriately move any open menus if the triggers are shifted in any way.

This PR also includes a change to the functions that are set initially if a pop-up is open by default. Previously the floating-ui positioner was activated, but the event emitter and the handler for mouse clicks outside of the menu were not, so some functionality was lost. This has been corrected in this PR by calling the `handleOpen()` function on `componentDidRender()`, as well as where it is run based on watching the `open` prop.

Finally, this PR also needed to address an issue with the way the `disableAutoUpdate` prop was being handled in order for the animation frame option from floating-ui to run on all instances of `rux-pop-up`. This also revealed an inherent issue with the fact that the `placement` prop is set default to auto, which is not an actual placement according to floating-ui, but an instruction on our end sent to the `autoPlacement` function to indicate that the pop-up itself should appear on whatever side is most appropriate given collision constraints. We allow users to have both `disableAutoUpdate` and placement='auto' (since having no placement _is_ placement='auto'). This is disingenuous to the user as they cannot have both no auto update, and a placement that updates automatically. This was addressed by setting the placement to 'top' if `disableAutoUpdate` is present without an explicit placement or 'auto'. This has been updated in Storybook.

## JIRA Link

[ASTRO-5071](https://rocketcom.atlassian.net/browse/ASTRO-5071)

## Motivation and Context

This change was brought to our attention by an internal Rocket employee working on a React application. Because of the project's classification, the exact circumstances cannot be replicated, but this fix should solve for the problem he encountered.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components/rux-pop-up/rux-pop-up.tsx
```diff
@@ -59,6 +59,11 @@ export class RuxPopUp {
      */
     @Prop({ reflect: true }) disableAutoUpdate: boolean = false
 
+    /**
+     * Turns on floatin-ui's autoUpdate animationFrame option to watch for trigger movements and replace the popup if movement is detected. Defaults to false.
```
markacianfrani, line 63:
```suggestion
     * watches for trigger movements and replace the popup if movement is detected. 
```

I think we should avoid calling out floatingui explicitly. as far as our end consumer is concerned, the only API they need to worry about is our rux pop up API. This also gives us the flexibility to replace floating-ui under the hood if we ever decided there was a strong reason to. 

also we dont need to specify here that it defaults to false. stencil will parse that out and storybook will show it under the default column in the API table

.changeset/ninety-crews-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrouxds/astro-web-components": minor
+---
+
+fix: Adding an intersection observer to watch the pop-up trigger so the menu moves with the trigger if it is pushed in any way.
```
markacianfrani, line 5:
```suggestion
Adds `enableAnimationFrame` prop to watch when the trigger moves
```

packages/web-components/src/components/rux-pop-up/rux-pop-up.tsx
```diff
@@ -289,6 +296,90 @@ export class RuxPopUp {
         }
     }
 
+    // private async _handleTriggerMovement() {
```
markacianfrani, line 299: can we remove this func now?

markacianfrani (approved):

PR #1072, opened 2023-02-14
fix(popup): fixed opened event firing 3 times

micahjones13:
## Brief Description

Fixes an issue where ruxpopup was firing the `ruxpopupopened` event three times when opened. 
This PR also updates snapshots. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-5614

## General Notes

This PR reverts a change that moved `handleOpen` to be on `componentDidRender` instead of `computePosition`, as this was leading to the multiple event emissions. 
`handleOutsideClick` event handling was updated in order to provide the functionality on a defaulted open pop-up. 
A new private variable named `_assignedOutsideClickEvent` which toggles a boolean if the `handleOutsideClick` listener has been added or not. 
The listener mentioned above will now be removed when the pop-up is closed, and re-added when it's open. 

## Motivation and Context

Stops multiple `ruxpopupopened` event emissions when pop-up opens.

## Issues and Limitations

The `startPositioner()` function now fires 3 times because it's been put under `componentDidRender`. I tried other lifecycle methods with no success because the popup needs to be fully rendered in order for the params on `startPositioner` to exist. I would like for this to not run more times than needed. The method at least does quick return if the popup isn't opened. Any suggestions there are much appreciated!

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components/rux-pop-up/rux-pop-up.tsx
```diff
@@ -133,6 +140,10 @@ export class RuxPopUp {
         this._handleTriggerKeyPress = this._handleTriggerKeyPress.bind(this)
         this._handleOutsideClick = this._handleOutsideClick.bind(this)
         this._setTriggerTabIndex = this._setTriggerTabIndex.bind(this)
+        if (this.open && !this._assignedOutsideClickEvent) {
+            window.addEventListener('mousedown', this._handleOutsideClick)
+            this._assignedOutsideClickEvent = true
+        }
```
micahjones13, line 145: @FMorrison87 adding this to the connectedCallback is how I got around the issue of the pop-up not closing on an off click if it was opened by default.

packages/web-components/src/components/rux-pop-up/rux-pop-up.tsx
```diff
@@ -141,7 +152,7 @@ export class RuxPopUp {
 
     componentDidRender() {
         if (this.open) {
-            this.handleOpen()
+            this._startPositioner()
```
micahjones13, line 154: This is a revert of the revert - handleOpen handles the emission of `ruxpopupopened`, so calling it here caused the extra events to be emitted. 
The reason this was changed to be handleOpen in a prev PR was so that the pop-up would show up in the correct spot and retain the ability to close on an off click if the pop-up was open by default. @FMorrison87 lemme know if that's incorrect

micahjones13:
@markacianfrani I don't think this needs a changeset because it fixes a bug that isn't yet published?

packages/web-components/src/components/rux-pop-up/rux-pop-up.tsx
FMorrison87: That's correct yeah.

FMorrison87 (approved):
Looks good! Good way to handle the use case where the pop-up wasn't setup to close if open by default and not emit more open events.

markacianfrani (approved):
looks good! this component could use a good refactor but we can tackle that separately. youre right, I dont think we need a changeset either

PR #1081, opened 2023-02-20
fix(focus) Remove :has selector from components that were previously using :focus-within

kiley-mitti:
## Brief Description

We had a couple of elements that were using :focus-within to display focus originally. However, that sometimes displayed focus when none was intended to be displayed.

So, in a previous quick-fix we changed the selector to :has(element:focus-within). That worked well, except that not all browsers support the :has selector yet. (Firefox does not, for example).

In this fix, I removed the :has selector from rux-segmented-button and rux-checkbox in favor of a small onFocus and onBlur script that programmatically adds a focused state to the elements that require focus styling. 

## JIRA Link

[ASTRO-5268](https://rocketcom.atlassian.net/browse/ASTRO-5268)

## General Notes

I had to apply similar but different solutions to checkbox and segmented button. The theory is the same but in segmented button I had to select the correct segment and programmatically add the --focused class. Checkbox has a 'isFocused' state that adds/removes the class on the checkbox.

**One question: I needed to add onFocus and onBlur to input fields on segmented button. Should I emit those events?** 

## Motivation and Context

This solution is compatible with all browsers.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/chatty-avocados-exist.md
+++ b/.changeset/chatty-avocados-exist.md
@@ -1,0 +1,8 @@
+---
+"@astrouxds/astro-web-components": patch
+"angular-workspace": patch
+"@astrouxds/angular": patch
+"@astrouxds/react": patch
+---
+
+feat(rux-checkbox) add ruxfocus event to rux-checkbox on focus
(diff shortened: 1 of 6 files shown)


micahjones13 (approved):
This seems really good to me, nice job! My vote is to add the focus/blur events to seg button for consistency sake, but we should talk with @markacianfrani about it.

markacianfrani (changes requested):
can we add tests for the new events? and separate changesets for the bugfix and the new events.

PR #1092, opened 2023-03-07
Mj datepicker playground

micahjones13:
### New Components 
- rux-datepicker
   - added tests
   - added SB page
- rux-calendar
  - Props
    - min -> the oldest date that the calendar will show
    - max -> the newest date the cal will show 
    - value -> The currently selected date on the calendar. 
    - date-in -> the default date used by calendar. When not provided, the calendar uses the current date. 
- Events 
   - `ruxdateselected` -> emitted when a selection is made on the calendar. Emits the selected date in ISO string format. 
   - Currently, only one day can be selected on the calendar at a time. Multi-select will come at a later time. 
   - added tests
      - VRT's use calendars that show the year from 2022. This is because the current day is denoted by a blue dot, and VRT's will fail every day since that dot will move. 
      - E2E tests include every month and leap years. 
   - added SB page
- rux-day (internal) 

### Modified Comps
- rux-input
    - Added a check to the `componentWillUpdate` to render a class based on if the input is being used in a datepicker comp or not. This is to hide the webkit calendar icon and replace it with our own. This is _hopefully_ a band-aid until I build out a custom input that supports Julian dates as well as Gregorian. 
 - rux-select 
    - Reflected the value prop on select to better support calendar since it uses two select menus 

### Misc
Updated the beta blurb on storybook to not burn user's eyes. Added github repo link for any feedback. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-5615

## Motivation and Context

Creates datepicker and calendar component. 

## Issues and Limitations

- Select menu's in calendar header 
   - The width of the month select is being set in a not-so-amazing way. This is because the designs call for the select menus to resize based on the selected option. This isn't easily do-able, so this was the simplest way. It is currently setting the width inline based on the selected month. 

### Needed Features 
- Date range selection 
- Min and max should not be limited to filtering years, it should also filter out months
   - Currently, only the year picker gets filtered by the min/max, even though the min/max takes in a full date, not just a year. In the future, we should filter the months as well. 
- In the datepicker, if you type a value into the input before selecting a day from the calendar, that day should be selected in the calendar when you open it. Currently only works from calendar -> input _not_ input -> calendar. 

- Julian (ordinal) date support will be added in a different PR. This one is massive enough as is. 
## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components/rux-calendar/rux-calendar.tsx
```diff
@@ -0,0 +1,547 @@
...
+            console.log('true')
+            console.log(this._maxDate, 'max date')
+            console.log(this._minDate, 'min date')
+            this._setStateWithDateIn()
+        }
+        this._handleYears(this._maxDate, this._minDate)
+    }
+    componentDidLoad() {
+        // attach event listeners to back and forward month arrows
+        this.el
```
FMorrison87, line 127: For getting the backward / forward elements, do you think it would be better to use the `ref` tagging pattern? That way you don't have to drill into the shadow, and potentially you could attach those listeners earlier.
This is the pattern I am referring to: setting a private variable `private _backArrow!: HTMLElement` and then setting the ref on the rux-icon in question to reference back to it: `ref={(el) => (this.backarrow = el!)}`. You should be able to use that as `this.backArrow.addEventListener('click', () => { my Func() })`

packages/web-components/src/components/rux-calendar/rux-calendar.tsx
```diff
@@ -0,0 +1,547 @@
...
+            .shadowRoot!.querySelector('#forward-month')
+            ?.addEventListener('click', () => {
+                this._handleForwardArrow()
+            })
+    }
+    //? Want to remove the selected prop from rux-day if changing month/year. Is this a good way to do it?
+    //? what kind of side effects can this have? What other times will componentWillUpdate fire?
+    //? Might become an issue when we use the month/year picker to change month/year?
+    componentWillUpdate() {
```
FMorrison87, line 141: Here is my suggestion: `componentWillUpdate() {
        const days = this.el.shadowRoot!.querySelectorAll('rux-day')
        days.forEach((day: HTMLRuxDayElement) => {
          if (day.selected) day.selected = false
        })
    }`

packages/web-components/src/components/rux-calendar/rux-calendar.tsx
```diff
@@ -0,0 +1,547 @@
...
+
+    /**
+     * Min date that the calendar will go to. Needs to be a valid date string.
+     */
+    @Prop() min?: string
+
+    /**
+     * Option to give the calendar a specfic month/year
+     */
+    @Prop({ attribute: 'date-in' }) dateIn?: string | number
```
FMorrison87, line 45: I think this should reflect so you can see the date-in attribute change in the DOM.

packages/web-components/src/components/rux-calendar/rux-calendar.tsx
micahjones13: I realized we don't even need to have a ref to the icons since they're being rendered always. I just threw the onClick onto them

FMorrison87: Even better!

packages/web-components/src/components/rux-button/readme.md
```diff
@@ -157,6 +158,7 @@ For more information about AstroUXDS usage outside of a Web Component environmen
 ```mermaid
 graph TD;
   rux-button --> rux-icon
+  rux-datepicker --> rux-button
```
micahjones13, line 161: ? wtf

packages/web-components/src/components/rux-input/rux-input.tsx
```diff
@@ -198,12 +200,27 @@ export class RuxInput implements FormFieldInterface {
     componentWillLoad() {
         this._handleSlotChange()
         this._setTogglePassword()
+
+        const shadowRootNode = this.el.getRootNode() as ShadowRoot
+        this._isDatePicker = shadowRootNode.host ? true : false
```
micahjones13, line 205: Determines if the input is being used in a datepicker or not. If it is, it adds a class that hides the webkit icon.

packages/web-components/src/components/rux-option/readme.md
```diff
@@ -20,6 +20,19 @@ onSlotchange doesn't fire.
...
+## Dependencies
+
+### Used by
+
+ - [rux-calendar](../rux-calendar)
+
+### Graph
+```mermaid
+graph TD;
+  rux-calendar --> rux-option
```
micahjones13, line 32: bro wtf

micahjones13: This is because calendar uses select which uses option.

packages/web-components/src/stories/time-region.stories.mdx
```diff
@@ -9,9 +9,9 @@ args.status.defaultValue = 'standby'
 export const StyledDiv = styled.div`
     position: relative;
     margin: 1rem 0;
-    border-left: 20px solid var(--status-symbol-color-fill-serious);
+    border-left: 20px solid var(--color-status-serious);
     background: white;
-    color: var(--status-symbol-color-fill-serious);
+    color: black;
```
markacianfrani, line 14: whats up with these story changes? is the branch just stale from main?

FMorrison87:
I found an interesting bug. If you select a day in the current month, then go back a few years, for example. If I select 9/17/23, then use the year selector and go back to 2020, the 17th is selected, but also another random day. If I go back further on the year selector, a third day is selected, if I pick a year further back than that, a fourth day. It's not super consistent, and the day selection appears random, but strange behavior.

packages/web-components/src/stories/time-region.stories.mdx
micahjones13: Prob some leftover crap from testing that I forgot to remove. Will revert that

markacianfrani:
closing but will keep branch for reference

PR #1121, opened 2023-04-27
fix(angular) fixed compile bug

kiley-mitti:
## Brief Description

rux-breadcrumb and rux-breadcrumb-item were not appearing in our angular module

## JIRA Link

[ASTRO-6131](https://rocketcom.atlassian.net/browse/ASTRO-6131)

## General Notes

It turned out that this had nothing to do with breadcrumb or breadcrumb -item at all. It was due to a race condition on compile that angular-proxies-list-ignore.ts was losing.

If proxies-list.ts didn't exist before running stencil's build, then the build would also fail. It turns out that this is because 'angular-proxies-list-ignore.ts' relied on the existence of proxies-list.ts to run. It assumed that the compiler had already generated the file by the time it ran. But sometimes this was not true, especially as the file got larger. I THINK what happened was that adding the two new components breadcrumb and breadcrumb-item pushed it over the edge.

So, in order to ensure that the proxies file WAS written before we added the ignore to the file, I created a small script that runs directly after stencil build and added it in the run order in package.json

## Motivation and Context

Solve the race condition and fix angular!

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


micahjones13 (approved):
🚀

markacianfrani (approved):

markacianfrani:
added a changeset and pruned some dead files

PR #1123, opened 2023-05-02
fix(tab): allow for dynamic selections of tab

micahjones13:
## Brief Description

This fixes a bug where programmatically changing the selected prop on an individual tab would not compute that tab's panel as selected/unselected. 
This [sandbox](https://codesandbox.io/s/sleepy-tristan-t9syno?file=/index.html) demo's the bug. You can see that the tab changes, but the content in the panel does not. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-6115

## General Notes

The solution in this PR is to add a `ruxtabselected` event to `rux-tab` that emits on an `@Watch('selected')`. This was chosen because we need someway to propagate the new selected info up to rux-tabs so that it can decide which panel should be shown. 

## Motivation and Context

Bug found by ACME team

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components/rux-tabs/rux-tab/rux-tab.tsx
```diff
@@ -28,13 +37,20 @@ export class RuxTab {
...
         if (this.el?.parentElement?.getAttributeNode('small')) {
             this.el.setAttribute('small', '')
         }
     }
 
+    @Watch('selected')
+    handleSelected() {
+        this.ruxTabSelected.emit(this.el)
+    }
+
```
micahjones13, line 53: Here's the new custom event that's emitted when selected changes.

packages/web-components/src/components/rux-tabs/test/tabs.spec.ts
```diff
@@ -297,7 +338,6 @@ test.describe('Multiple tabs on same page', () => {
         await expect(topContent2).toBeVisible()
         await expect(bottomContent).toBeVisible()
     })
-    //TODO: Replace instances of e.evaulate here with the .hasAttribute
```
micahjones13, line 300: Refactored tests while I was in there

FMorrison87: Good call on the test refactoring. Not having to use that evaluate stuff to check for attributes is great!

FMorrison87 (approved):
I think this is a good solution. Adding a new custom event, emitting it and listening for it should allow users some good flexibility. Nice fix.

packages/web-components/src/components/rux-tabs/rux-tabs.tsx
```diff
@@ -36,6 +36,24 @@ export class RuxTabs {
      */
     @Prop() small?: boolean
 
+    // This allows us to hear the selected prop change on tab.
+    // Once we hear it, we need to update the related panels visibilty accordingly.
+    @Listen('ruxtabselected', { target: 'window' })
+    handleTabselected(e: CustomEvent) {
+        console.log('heard ruxtabselected event')
```
markacianfrani, line 43:
```suggestion
      
```

packages/web-components/src/components/rux-tabs/rux-tab/rux-tab.tsx
```diff
@@ -28,13 +37,20 @@ export class RuxTab {
 
     @Element() el!: HTMLRuxTabElement
```
markacianfrani, line 39:
```suggestion
   /**
     * Fired when tab is selected
     */
```

markacianfrani (approved):
looks good! some cleanup suggestions added, just needs changeset and whatevers going on w/ the tests failing?

packages/web-components/src/components/rux-tabs/rux-tab/rux-tab.tsx
```diff
@@ -28,13 +37,23 @@ export class RuxTab {
...
     connectedCallback() {
         //handle small on init
         if (this.el?.parentElement?.getAttributeNode('small')) {
             this.el.setAttribute('small', '')
         }
     }
 
+    @Watch('selected')
+    handleSelected() {
+        if (this.selected) this.ruxTabSelected.emit(this.el)
```
micahjones13, line 54: @markacianfrani I updated this to fire only when selected is changed to be true, otherwise whenever it changes it'll fire this event which doesn't make sense.

packages/web-components/src/components/rux-tabs/rux-tabs.tsx
```diff
@@ -36,6 +36,25 @@ export class RuxTabs {
...
     @Prop() small?: boolean
 
+    // This allows us to hear the selected prop change on tab.
+    // Once we hear it, we need to update the related panels visibilty accordingly.
+    @Listen('ruxtabselected', { target: 'window' })
+    handleTabselected(e: CustomEvent) {
+        const target = e.target as HTMLRuxTabElement
+        //* only change the classlist of panels assoiciated with this rux-tabs component
+        const children = Array.from(this.el.children)
+        if (target.selected && children.includes(target)) {
```
micahjones13, line 46: @markacianfrani @FMorrison87 The reason the tests were failing is because I didn't filter out the events to be just the ones from children of the current rux-tabs, so if you had more than one rux-tabs on a page, it'd get jumbled together with the events from other comps. Thank god for tests!

PR #1135, opened 2023-05-15
fix(rux-segmented-button) programmatically select segments

kiley-mitti:
## Brief Description

Segmented button had an issue when trying to programmatically select a segment using rux-segmented-button's 'selected' attribute. It would select any item that was not initially selected.

## JIRA Link

[ASTRO-6212](https://rocketcom.atlassian.net/browse/ASTRO-6212)

## General Notes

I put a listener on 'selected' and updated the input check when it changes.

## Motivation and Context

This use case is likely small but it was making the component act strangely with the controls in Storybook and Playground.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-segmented-button/test/segmented-button.spec.ts
+++ b/packages/web-components/src/components/rux-segmented-button/test/segmented-button.spec.ts
@@ -113,6 +113,41 @@ test.describe('Segmented-button', () => {
         await expect(changeEvent).toHaveReceivedEventTimes(1)
         await expect(changeEvent).toHaveReceivedEventDetail('Second segment')
     })
+    test('it properly reselects a pre-selected segment', async ({ page }) => {
+        const template = `
+        <div style="padding: 2.5% 5%">
+            <rux-segmented-button></rux-segmented-button>
+        </div>
+        `
+        await page.setContent(template)
+        await page.addScriptTag({
+            //first is selected by default
+            content: `
+            const segmented = document.querySelector('rux-segmented-button')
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/rux-segmented-button/test/segmented-button.spec.ts
```diff
@@ -113,6 +113,41 @@ test.describe('Segmented-button', () => {
         await expect(changeEvent).toHaveReceivedEventTimes(1)
         await expect(changeEvent).toHaveReceivedEventDetail('Second segment')
     })
+    test('it properly reselects a pre-selected segment', async ({ page }) => {
+        const template = `
```
markacianfrani, line 117: Because this one's a smol change, try commenting out the change you made. The test should fail. When I comment it out, the test still passes. Can we think of a different approach here to capture the issue we're seeing?

markacianfrani (changes requested):

packages/web-components/src/components/rux-segmented-button/test/segmented-button.spec.ts
kiley-mitti: Playwright not working how I thought it was. This should be solved :)
I added some evaluated to grab the properties and then added expect(evaluation).toBe(true) and it was happy. with that.

markacianfrani (approved):
looks good! left a suggestion in the test

PR #1144, opened 2023-05-24
fix(progress): respect max when value increases

micahjones13:
## Brief Description

Changes progress to respect max instead of changing max to equal value when value gets beyond max. 
This seems to be legacy functionality and isn't justified anymore. Was brought up in the demo day and was declared a bug. 
This fix swaps the (backwards) functionality of max changing to equal value with value instead changing to equal max if value is being incremented past the max prop. 

Updates a test.

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-6244

## Motivation and Context

Noticed by user. Makes progress act as you would expect 

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-progress/rux-progress.tsx
+++ b/packages/web-components/src/components/rux-progress/rux-progress.tsx
@@ -13,11 +13,11 @@ export class RuxProgress {
     /**
      * Current progress value between 0 and 100 (or the max, if defined below).
      */
-    @Prop() value?: number = 0
+    @Prop({ mutable: true }) value?: number = 0
     /**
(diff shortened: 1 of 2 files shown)


kiley-mitti (approved):
not blocking, but do we want the value to decrease with max-value if they're both at max-value?

example: 100/100 decreases the max to 90
100/90 or 90/90? the next time one touches the value it will change to 90/90 but not before then.

markacianfrani:
good question. that makes sense to me but we should see what ux thinks. and if we did make that the default behavior, would someone be able to opt out of it if they needed?

PR #1148, opened 2023-06-01
Mj.dual range slider

micahjones13:
## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-slider/rux-slider.tsx
+++ b/packages/web-components/src/components/rux-slider/rux-slider.tsx
@@ -57,8 +57,9 @@ export class RuxSlider implements FormFieldInterface {
      * Current value of the slider. The default value is halfway between the specified minimum and maximum. - [HTMLElement/input_type_range>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range)
      */
 
-    @Prop({ mutable: true }) value: number =
-        (this.max! - this.min!) / 2 + this.min!
+    @Prop({ mutable: true }) value: number = this.endVal
+        ? this.endVal
+        : (this.max! - this.min!) / 2 + this.min!
 
     /**
      *  Shows tick marks and labels in the order provided and aligns evenly based on the length.
@@ -95,15 +96,14 @@ export class RuxSlider implements FormFieldInterface {
     @Prop({ attribute: 'error-text' }) errorText?: string
 
     /**
-     * Enables a dual-range slider
+     * The value of the first thumb if using a dual-range slider
      */
-    @Prop() range?: boolean = false
+    @Prop({ attribute: 'start-val', mutable: true }) startVal?: number
 
     /**
      * The value of the second thumb if using a dual-range slider
      */
-    @Prop({ attribute: 'dual-value', mutable: true }) dualValue?: number
-
+    @Prop({ attribute: 'end-val', mutable: true }) endVal?: number
     /**
      * Fired when the value of the input changes - [HTMLElement/input_event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event)
      */
@@ -119,14 +119,14 @@ export class RuxSlider implements FormFieldInterface {
 
     componentWillLoad() {
         this._updateValue()
-        this._updateDualValue()
+        this._updateStartVal()
         this._getBrowser(navigator.userAgent.toLowerCase())
         this._handleSlotChange()
     }
 
     connectedCallback() {
         this._onInput = this._onInput.bind(this)
-        this._onDualInput = this._onDualInput.bind(this)
+        this._onStartValInput = this._onStartValInput.bind(this)
         this._onBlur = this._onBlur.bind(this)
         this._handleSlotChange = this._handleSlotChange.bind(this)
         this._onChange = this._onChange.bind(this)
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/rux-slider/rux-slider.tsx
```diff
@@ -157,46 +184,69 @@ export class RuxSlider implements FormFieldInterface {
...
+    }
+
+    private _onMinValInput(e: Event) {
+        const target = e.target as HTMLInputElement
+        this.minVal = parseFloat(target.value)
+        if (this.minVal >= this.maxVal! && this.strict) {
+            this.minVal = this.maxVal
+            //? Still not entirely sure why we have to do this.
+            target.value = this.minVal!.toString()
+        }
```
micahjones13, line 249: @markacianfrani relevant funcs for the implementation of the strict prop

packages/web-components/src/components/rux-slider/rux-slider.tsx
```diff
@@ -157,46 +178,66 @@ export class RuxSlider implements FormFieldInterface {
...
+        if (this.value !== undefined) {
+            this.value = parseFloat(target.value)
+            if (this.value <= this.minVal! && this.strict) {
+                this.value = this.minVal!
+                // this.value = this.maxVal!
+                //? Still not entirely sure why we have to do this.
+                target.value = this.value.toString()
+            }
+        } else {
+            this.value = parseFloat(target.value)
```
FMorrison87, line 227: Do you need this else clause? It looks like you're setting `this.value` as a float regardless. Can you bring back line 199 where you set it before hitting the first if statement?

FMorrison87 (commented):
Left a few comments, final one is, are there supposed to be VRT snaps from progress bar in this PR?

packages/web-components/src/stories/slider.stories.mdx
```diff
@@ -219,13 +219,68 @@ export const WithAxisLabels = (args, data) => {
...
+be evenly distributed across the progress bar.
 
 #### Ticks Only
 
 If you only want to show tick marks without labels, make use of the `ticks-only` property. NOTE: you will still need to pass in the array of strings to determine how many
 ticks to show.
 
+### Dual Range
+
+Adding the `min-val` property to the slider will add a second thumb, creating a dual-range slider.
```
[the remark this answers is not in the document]
micahjones13: That's a good point, I always get those mixed up. I'll change the reference to be attribute since what I'm getting at is how the user would use the attribute

packages/web-components/src/components/rux-slider/rux-slider.tsx
micahjones13: nope, that's extra

packages/web-components/src/utils/utils.ts
```diff
@@ -58,14 +58,15 @@ export const renderHiddenInput = (
     name: string,
     value: string | undefined | null,
     disabled: boolean,
-    checked?: boolean
+    checked?: boolean,
+    secondInput?: boolean
 ) => {
     let input = container.querySelector(
         'input.aux-input'
```
FMorrison87, line 65: did you mean `input.rux-input`?

packages/web-components/src/utils/utils.ts
```diff
@@ -58,14 +58,15 @@ export const renderHiddenInput = (
...
     ) as HTMLInputElement | null
     if (always || hasShadowDom(container)) {
         if (checked || checked == undefined) {
-            if (!input) {
+            if (!input || (secondInput && input)) {
                 input = container.ownerDocument!.createElement('input')
                 input.type = 'hidden'
                 input.classList.add('aux-input')
```
FMorrison87, line 72: `aux-input`?

FMorrison87 (changes requested):

packages/web-components/src/utils/utils.ts
micahjones13: No, it's `aux-input`. It's a hidden, native input element that just facilitates form submissions since shadow-dom has probs with that.

kiley-mitti (changes requested):
It looks awesome! I know you're peeking at making it work with step. If you just hit rerequest once you've got that how you want it I'll come in and check it post haste! :)

FMorrison87 (commented):
Looks great! Needs the test for the strict prop and for clicking right of the right thumb, then it looks good to go!

micahjones13:
> Looks great! Needs the test for the strict prop and for clicking right of the right thumb, then it looks good to go!

I can't for the life of me get the click to work in testing for moving the right thumb. Using the same method as I did to move the left thumb, but as soon as position y gets over 20 it breaks. I tried `force: true` but still no luck. If anyone has better ideas, plz let me know

kiley-mitti (approved):
Beautiful! Needs a changes but looks amazing. Great job Micah! It's brilliant. :)

FMorrison87 (approved):

FMorrison87:
I was testing programmatically updating the minVal and value prop and I noticed that the way the thumbs are set in the test file is with an attribute called `max-val`. Since this is not a prop when I update the component using the minVal and value props I get reflection on the min-val attribute, but the max-val attribute does not reflect the new value. Should it?

micahjones13:
> I was testing programmatically updating the minVal and value prop and I noticed that the way the thumbs are set in the test file is with an attribute called `max-val`. Since this is not a prop when I update the component using the minVal and value props I get reflection on the min-val attribute, but the max-val attribute does not reflect the new value. Should it?

ah great catch, that's a vestige of the first implementation. I'll fix that right now. Thank you!!

FMorrison87 (approved):

markacianfrani (approved):

PR #1151, opened 2023-06-02
fix(rux-status) center icon within its space by default

kiley-mitti:
## Brief Description

I made a couple of small css tweaks to rux-status:
1) rux-status now displays inline-block rather than block
2) ~~::part[status] now has margin-inline(left and right) auto to center it in the space it takes up.~~
**Edit:** Removed 2 in a subsequent commit after speaking with mark.

## JIRA Link

[ASTRO-6247](https://rocketcom.atlassian.net/browse/ASTRO-6247)

## General Notes

This MIGHT be a major change or possibly minor. it is not just a patch since it changes the way the dimensions of this component work.. but it is easily changed back to the way it was before with a small css tweak.

## Motivation and Context

This was a pain point for the TTC team and I think it's a legitimate one. In an application they needed to put status in .. basically its own table cell.

- Because rux-status defaults to block it took up the whole table cell.
- Because the svg icon wasn't centered in rux-status, it aligned itself left.
- Because rux-status  was taking up the WHOLE CELL, setting center on the cell didn't center the icon.

Example: [https://codesandbox.io/s/agitated-hertz-7evbhb?file=/index.html](https://codesandbox.io/s/agitated-hertz-7evbhb?file=/index.html)

This is totally an easy fix, just add `margin: 0 auto` to `part=status`
But.. this is SUCH a common use case that we even center the icon in our storybook examples.
[Storybook](https://astro-components.netlify.app/?path=/docs/components-status--all-variants)

And the only reason that the icon in he second set of examples (the one with all icons) centers without `margin: auto` is because `display:flex` is taking over the `display:block` on rux-status and forcing it to only be the width of its content. 

Anyway I thought that such its such a simple change having a sample of it up for discussion might be good. So this draft is that sample. :)

## Issues and Limitations

This is a breaking change. how severe is debatable but if we want to be cautious its probably major.. but its only major for people who, for whatever reason, WANT rux-status to take up all the space it is given while also leaving the icon to the left.. I can't say how many people that is but I don't know that its many.

If we remove change #1 (changing from block to inline-block) then it is either a patch or minor change since we're not changing the dimensions of the component.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [x] Breaking change

## Checklist


markacianfrani:
going to close this for now due to breaking change philosophy. tickets have been updated so we can slot it into v8 since its not a huge change

PR #1169, opened 2023-06-28
feat(monitoring-icon) add 'size' prop

kiley-mitti:
## Brief Description

This adds a size prop to the monitoring icon which gets passed down to the rux-icon inside.

## JIRA Link

No ticket, just a suggestion.

## Related Issue

[976](https://github.com/RocketCommunicationsInc/astro/discussions/976)

## Motivation and Context

This has been a requested feature over time which we have put off because Design said we would run into legibility issues. Then in the Collab a couple of weeks ago, Jean said something about needing to resize the monitoring icon to fit in a shrunk down version of the GSB and design agreed that that customization seemed necessary.

So.. taking that and running with it, ALL this change does is allow you to change the size of the icon. It does not change the way the monitoring icon component works in any other way.

## Issues and Limitations

This means that if someone sets the icon to be too large (over 4.5rem) or too small, it looks odd and they would have to further modify some styles. This can be fixed by modifying the way that rux-monitoring-icon is styled but it would be a breaking change.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components.d.ts
```diff
@@ -12088,6 +12088,10 @@ export namespace Components {
           * If provided and greater than `0`, displays an outlined number badge at the bottom right of the icon. Numbers above `9999` are abbreviated to `'10K'` or `'100K'` for numbers in the thousands, `'1.5M'` for millions, `'1.5B'` for billions, and uses `'∞'` for one trillion or higher.
          */
         "notifications": number;
+        /**
+          * The size of a chosen Astro icon. Can be 'extra-small', 'small', 'normal', 'large', 'auto' or any custom value ('30px', '1rem', '3.321em')
+         */
+        "size": string;
```
markacianfrani, line 12094: maybe rename to `icon-size` instead? at first glance, I would think `size` would refer to the host element not the icon within. 0/10

markacianfrani (approved):

PR #1188, opened 2023-07-20
fix(tooltip): width bug

micahjones13:
## Brief Description

removes `display: inline-flex` from the `:host` of `rux-tooltip`. 
This causes the width of anything wrapped in the tooltip to not properly inherit. However, it also means that sometimes the tooltip will get wrapped to a new line when used in the default `auto` placement. Changing this placement to `right` should fix it, though. 

Bug codepen: https://codepen.io/micahjones/pen/LYXrdOo
## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-5643

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


kiley-mitti (approved):
I think this is a fine solution for reasons I put in the thread you linked, Micah. I am NOT sure whether it is breaking or not though. It might spontaneously change tooltip widths that people have set up. Maybe minor?

PR #1191, opened 2023-07-21
Changed rux-select to use box-shadow rather than border

nortonprojects:
## Brief Description

`rux-input` uses box-shadow, not border, which gives it a height of 36px by default. `rux-select` uses the opposite, which gives it a height of 38px by default. This makes aligning a `rux-select` and `rux-input` extremely frustrating.

This PR just changes `rux-select` to be consistent with `rux-input`, making the height the same and alignment trivial.

## Types of changes

- [X] Bug fix
- [ ] New feature
- [X??] Breaking change

## Checklist

- [ ] This PR adds or removes a Storybook story.
- [ ] I have added tests to cover my changes.
- [ ] Regressions are passing and/or failures are documented
- [ ] Changes have been checked in evergreen browsers

PR #1209, opened 2023-08-09
fix(rux-select) make height align with rux-input height

kiley-mitti:
## Brief Description

This is a fix for the problem @nortonprojects brought up here:
https://github.com/RocketCommunicationsInc/astro/pull/1191

It takes rux-select and makes its height the same as rux-input so that the two can be easily aligned. It's actually a fix to rux-select since in figma the heights were always supposed to be as they are in this PR.

## General Notes

Jacob's proposed solution was to turn select's border into a box shadow which would make it match rux-input. This is a great solution and consistency is great! However, I'd actually like to change rux-input so that its border is actually a border and not a box shadow. I think that moving the fake border from box-shadow to a real border will solve other paper cuts in development. To that end, this is the solution I propose for the rest as well. I've done a calc on padding so that it takes border into account.

## Motivation and Context

Forms are annoying to align when the inputs are different sizes. This fixes that for select/input.

## Issues and Limitations

Might be a breaking change for people who have already tried to solve this manually by changing aux-input's size rather than select's.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [x] Breaking change

## Checklist


packages/web-components/src/index.html
```diff
@@ -16,7 +16,468 @@
...
+        </section>
+        <section class="columns">
+            <rux-input value="Hi there!" label="With Label"></rux-input>
+            <rux-select label="With Label">
+                <rux-option label="Option Option Option"></rux-option>
+                <rux-option label="Option Option Option"></rux-option>
+                <rux-option label="Option Option Option"></rux-option>
+            </rux-select>
+        </section>
+    </body>
```
markacianfrani, line 482:
```suggestion
    <body></body>
```
clean body

markacianfrani (approved):

PR #1211, opened 2023-08-14
fix(rux-status) correct status symbol sizes

kiley-mitti:
## Brief Description

During a TTC app review, we discovered that the svgs that power rux-status were the wrong dimensions. With Jason's help I corrected them. This does not change the size of rux-status itself, only minority affects the visuals so this change should be non breaking.

## JIRA Link

[ASTRO-6719](https://rocketcom.atlassian.net/browse/ASTRO-6719)

## Motivation and Context

The status symbols were slightly smaller than intended.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


.changeset/fluffy-bats-tie.md
```diff
@@ -0,0 +1,11 @@
+---
+"@astrouxds/astro-web-components": patch
+"angular-workspace": patch
+"@astrouxds/angular": patch
+"astro-angular": patch
```
markacianfrani, line 5: @micahjones13 does it matter if we leave this in for release? the framework tests dont need to be included but Im not sure if it actually matters?

micahjones13: I don't think so. I've been removing them since it hasn't been too cumbersome, but even if we left them in I'm p sure it wouldn't matter at all

packages/web-components/src/index.html
```diff
@@ -16,7 +16,36 @@
...
+        </section>
+        <section class="light-theme">
+            <rux-status status="off"></rux-status>
+            <rux-status status="standby"></rux-status>
+            <rux-status status="normal"></rux-status>
+            <rux-status status="caution"></rux-status>
+            <rux-status status="serious"></rux-status>
+            <rux-status status="critical"></rux-status>
+        </section>
+    </body>
```
markacianfrani, line 50:
```suggestion
    <body>
    </body>
```

markacianfrani (commented):
curious question about tree node VRT, clean body. can merge this though if blocking something else.

packages/web-components/vrt-snaps/rux-tree-node/test/basic/tree-node.vrt.spec.ts-snapshots/Tree-Node-has-no-visual-regression-vrt-1-chromium-vrt-light-linux.png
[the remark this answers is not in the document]
kiley-mitti: I cleaned the index and also reverted that image. I don't know why it changed and I must have missed it because it fell in with the 'tree' changes which I knew would change because they have status icons in them. :) Anyway I'm curious if it passes now too. Looks like the font was different. Maybe it took the snap before docker caught up and put the correct font in place? 🤷‍♀️

packages/web-components/src/index.html
```diff
@@ -16,7 +16,20 @@
...
+                padding: 2rem;
+            }
+            section {
+                display: flex;
+                gap: 8px;
+            }
+            .light-theme {
+                background-color: #fff;
+            }
+        </style>
```
markacianfrani, line 30:
```suggestion
```

markacianfrani (approved):

Issue #1216, opened 2023-08-16
`<RuxTooltip />` renders block-level children taller than inline children


jm-bh:
**Describe the bug**
Block-level elements inside a `<RuxTooltip />` make the tooltip taller than it should be.

**To Reproduce**

1. Create a `<RuxTooltip />`
2. Add a block-level element
3. Observe incorrect height

**Current behavior**
The tooltip is significantly taller than its children.

**Expected behavior**
The tooltip should not be taller than its children. (Inline elements still show the correct height.)

**Screenshots**
![image](https://github.com/RocketCommunicationsInc/astro/assets/138606233/653d04d8-ec1a-4c22-beed-84f36b0b26b3)
![image](https://github.com/RocketCommunicationsInc/astro/assets/138606233/81b80029-8376-40c8-82a0-b1a9b1ab81b4)

**CodeSandbox**
[StackBlitz](https://stackblitz.com/edit/astro-nextjs-repro?file=components%2FTable.tsx) (`/components/table`, line 72)

**Environment (please complete the following information):**

- Browser [e.g. chrome, safari]: **Chrome 115.0.5790.171**
- @astrouxds/astro-web-components version [e.g. 7.0.0]: **7.15.2**
- Framework, if applicable (React, Angular, Vue, Svelte, ect): **React**
- Framework wrapper version, if applicable (@astrouxds/react or @astrouxds/angular): **`@astrouxds/react`**

**Additional context**
Possibly related to #1188

micahjones13:
Hey @jm-bh,
This issue has been fixed and will be included in today's release (7.16.0). Thanks for raising this!

PR #1217, opened 2023-08-16
fix(rux-tooltip)tooltip is a different size when trigger is a block level element vs inline

kiley-mitti:
## Brief Description

I suspect that this was because there was an invisible <Fragment> in the component. I'm not sure why it was there so I'm hesitant to call this a fix without more testing.. but removing it and rearranging the code did seem to fix this specific issue. I'm putting this in draft so that Micah can have a chance to mess with it too.

## JIRA Link

[ASTRO-6764](https://rocketcom.atlassian.net/browse/ASTRO-6764)

## Related Issue

Brought to our attention by this issue: [#1216](https://github.com/RocketCommunicationsInc/astro/issues/1216) (Thank you!)

## General Notes

This does break the VRTs for tooltips but I think that's actually ok. They are all using <div class="trigger"> so they had the extra space built in.

## Motivation and Context

Makes tooltip behave in a more intuitive way when faces with interior elements.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components/rux-tooltip/rux-tooltip.tsx
```diff
@@ -264,9 +263,8 @@ export class RuxTooltip {
                         ref={(el) => (this.trigger = el!)}
                         aria-describedby="tooltip"
                     >
-                        <slot onSlotchange={_handleSlotChange} />
+                        <slot onSlotchange={_handleSlotChange}></slot>
                     </span>
-                    <Fragment>&#8203;</Fragment>
```
kiley-mitti, line 269: I'm not sure what this was here for but it does seem to be the problem child.

markacianfrani: I'm pretty sure this was to solve this issue:

```
whatup its <rux-tooltip message="mark">yer</rux-tooltip> boi
```
Screenshot 2023-08-16 at 4 42 46 PM

If I double click on "yer" to select the text, the tooltip text should not be selected. 
https://codepen.io/rocketmark/pen/eYbOrab

but in your PR with this out, this is what you get. 
Screenshot 2023-08-16 at 4 43 32 PM

kiley-mitti: Oh good that helps! I'll see if I can solve for both in the morning. Thanks for the context!!

kiley-mitti: @markacianfrani I'm pretty sure that we don't want people to accidentally select the tooltip at all, so what if, as an alternate solution, we just add the css `user-select` (and `-webkit-user-select`) `none`?

I've updated the PR with this. this should make it so that not only does it not select the tooltip when double clicking but it also won't select it if you drag to select the whole phrase.

markacianfrani (approved):

PR #1224, opened 2023-09-12
feat(dialog): toggleOpen, openDialog, closeDialog methods

micahjones13:
## Brief Description

This adds a public `toggleOpen` method to rux-dialog that makes the dialog easier to open and close. 

## JIRA Link

https://rocketcom.atlassian.net/browse/ASTRO-6841

## Motivation and Context

On FDS we were having a problem where the open prop on dialog would get out of sync, and you had to listen to the closed and opened events in order to toggle the open prop correctly, which can be cumbersome and unintuitive. 

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-dialog/rux-dialog.tsx
+++ b/packages/web-components/src/components/rux-dialog/rux-dialog.tsx
@@ -82,7 +82,7 @@ export class RuxDialog {
      * Toggles the dialog's open prop.
      */
     @Method()
-    async toggleOpen() {
+    async toggle() {
         this.open = !this.open
     }
 
@@ -90,7 +90,7 @@ export class RuxDialog {
      * Opens the dialog
      */
     @Method()
-    async openDialog() {
+    async show() {
         this.open = true
     }
 
@@ -98,7 +98,7 @@ export class RuxDialog {
      * Closes the dialog
      */
     @Method()
-    async closeDialog() {
+    async hide() {
         this.open = false
     }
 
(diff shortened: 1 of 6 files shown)


packages/web-components/src/components/rux-dialog/rux-dialog.tsx
```diff
@@ -77,6 +78,11 @@ export class RuxDialog {
     })
     ruxDialogClosed!: EventEmitter<boolean | null>
 
+    @Method()
+    async toggleOpen() {
+        this.open = !this.open
+    }
+
```
micahjones13, line 85: Is there any benefit to returning `this.open`, or should this just remain a void function?

Paterma (approved):
🎉 🎉

packages/web-components/src/components/rux-dialog/rux-dialog.tsx
kiley-mitti: Probably can stay void. Can get this.open from the open attribute on the dialog

kiley-mitti (changes requested):
Just a note for storybook purposes. This is pretty straight forward. I THINK the method can return void.. but looking at popup its hide and show methods do return false and true respectively, so maybe for consistency it should? 1/10

kiley-mitti (approved):

markacianfrani (approved):

PR #1227, opened 2023-09-18
fix(rux-segmented-button)

kiley-mitti:
## Brief Description

`rux-segmented-button` is radio group under the hood. Because all of the segmented button groups were named the same thing, there were issues tabbing between them. This fix adds 5 random numbers to the name of the underlying inputs so that when two button sets are next to each other, you can tab between them as expected.

## JIRA Link

[ASTRO-5270](https://rocketcom.atlassian.net/browse/ASTRO-5270)

## Motivation and Context

This isn't a huge gamebreaking issue.. especially since segmented buttons are suggested only to be used in certain situations. But keyboard controls not functioning as one would expect can be hugely annoying so its worth shoring up anyways.

## Issues and Limitations

Random numbers are generally not a great idea but in this specific instance I think its alright since no one can access the 'name' property on input anyway.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-segmented-button/rux-segmented-button.tsx
+++ b/packages/web-components/src/components/rux-segmented-button/rux-segmented-button.tsx
@@ -127,6 +127,10 @@ export class RuxSegmentedButton {
         target.closest('li')?.classList.remove('--focused')
     }
 
+    //this random number is appended to <input name="" in order to give each segmented button group a unique name
+    //this allows users to tab between button groups that are adjacent to each other
+    private random = Math.floor(Math.random() * 100000)
+
     render() {
         return (
             <ul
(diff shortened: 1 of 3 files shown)


markacianfrani (approved):
might also be able to make use of the [autoincrement id pattern](https://github.com/RocketCommunicationsInc/astro/blob/main/packages/web-components/src/components/rux-push-button/rux-push-button.tsx#L23) we use in a few other places when the element is created instead of random but that works

packages/web-components/src/components/rux-segmented-button/rux-segmented-button.tsx
```diff
@@ -127,6 +127,10 @@ export class RuxSegmentedButton {
         target.closest('li')?.classList.remove('--focused')
     }
 
+    //this random number is appended to <input name="" in order to give each segmented button group a unique name
+    //this allows users to tab between button groups that are adjacent to each other
+    private random = Math.floor(Math.random() * 100000)
```
micahjones13, line 132: Is there a chance that this same random number could be given to another name attr? I mean I think the chances are 1 in a 100000. I'm more asking out of curiosity lol

micahjones13 (approved):
I think we should use the auto-increment pattern that Mark mentioned instead of random nums.

PR #1249, opened 2023-10-25
feat(selcet): reflect value on select

micahjones13:
## Brief Description

This PR adds a reflect to the value prop of rux-select, as well as an initValue function that runs on connected callback. 
This function selects a the first 'rux-option' as the value if a value is not already defined. 

I ran into an issue using FormData in a vanilla setting where select menu wouldn't have a value if you didn't interact with it. This could mean that form submissions could be incorrect if the dev used a select menu without a 'select an option...' option.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/web-components/src/components/rux-select/rux-select.tsx
+++ b/packages/web-components/src/components/rux-select/rux-select.tsx
@@ -84,7 +84,7 @@ export class RuxSelect implements FormFieldInterface {
     /**
      * The value of the selected option. If multiple is true, this is an array.
      */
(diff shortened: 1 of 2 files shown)


markacianfrani (changes requested):
I dont think we should reflect an array. did you try it with an excessive number of selected options w/ multi? 

https://web.dev/articles/custom-elements-best-practices#aim-to-only-accept-rich-data-objects,-arrays-as-properties.

PR #1269, opened 2023-12-01
Km.6882 add keyboard control tooltip

kiley-mitti:
## Brief Description

Added an escape key listener that will close a tooltip if the tooltip trigger is either hovered or focused. If it is neither the escape key won't act on it.

## JIRA Link

[ASTRO-6882](https://rocketcom.atlassian.net/browse/ASTRO-6882)

## Motivation and Context

Cory (from Shoelace) mentioned that tooltips should have ability to close without moving the mouse or unfocusing. Seemed like a good idea.

## Issues and Limitations

Make sure this is the functionality we want. I could remove the if statements in the listener and have all tooltips close whenever escape is pressed. That would also be valid. This is just more specific.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/rich-cherries-hunt.md
+++ b/.changeset/rich-cherries-hunt.md
@@ -1,0 +1,5 @@
+---
+"@astrouxds/astro-web-components": patch
+---
+
+Tooltip - allow pressing escape to close the tooltip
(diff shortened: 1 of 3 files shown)


kiley-mitti:
test isn't passing for some reason I'll get it going and add a changeset asap

micahjones13 (approved):

PR #1274, opened 2024-01-09
Km.24 hour input exploration

kiley-mitti:
## Brief Description

--WIP --

This is a proof of concept for a new Astro Component for forces 24hour/12hour time input. It uses iMaskJs for masking.

The component is basically complete with just review and bug reports to work through, however, this PR also needs cleaning up and storybook additions before committing.

Update 1/22 - removed the 12hour masking functionality in favor of `type="time"` when 24h is not chosen  

TODO:
- should changing include-seconds or timeformat programmatically update the mask? - probably.
 - add storybook
- add changeset

## JIRA Link

[R2C2-119](https://rocketcom.atlassian.net/jira/software/projects/R2C2/issues/R2C2-119)

## Motivation and Context

This exploration of new components was undertaken because `<rux-input type="time">` is not as flexible as it could be regarding time input. There have been several requests from members of the current project for a component capable of forcing 24h time input.

## Issues and Limitations

There are still a couple of bugs I'd like to sort out at this time:
- adding `include-seconds` / changing `timeformat` after first render will not actually change the input masking. This should be handled.
- iMaskJs is a little large as a dependency.. also, not positive all spaces will approve its use. Ultimately we should write our own masking but it will take more time than this POC.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


markacianfrani (approved):
LGTM

PR #1309, opened 2024-04-12
Add status and icon resize to Toast

micahjones13:
## Brief Description

Adds the ability to assign a status to rux-toast via the new `status` prop. This styles the toast with the corresponding status symbol, border color and icon color. 

Adds the ability to resize the close icon via a new `--iconSize` CSS custom prop. For example: 

```
rux-toast {
    --iconSize: 24px;
}
```

## Related Issue
https://github.com/RocketCommunicationsInc/astro/discussions/1308
## Motivation and Context

There is a use case where different types of toasts should be displayed as a user uses an application.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


kiley-mitti (approved):
Looks great! Non-blocking but maybe a storybook story should be added to display the new status prop.

Issue #1340, opened 2024-09-16
Cursor appears to be inside input, when it is technically not.


danielwlemmon:
**Describe the bug**
When a user clicks on closer to the edge of an input a cursor is placed just outside of the actual input element (at the span?) making it look like the user can enter text, but not characters show.
![input cursor](https://github.com/user-attachments/assets/3e409ad8-a4d7-4ae5-bff9-f9386dd4f668)

**To Reproduce**

1. Click on close to the beginning or end of an input and try to type.
2. See screen grab above.

**Current behavior**
A clear and concise description of what is currently happening.
A rux-input becomes something like this at render
<div part="input-field" class="rux-input rux-input--medium rux-input--invalid">
  <span part="prefix" class="rux-input-prefix"><slot name="prefix"></slot></span
  ><input
    name=""
    type="text"
    aria-invalid="true"
    class="native-input"
    id="rux-input-16"
    spellcheck="false"
    part="input"
    autocomplete="off"
  /><span part="suffix" class="rux-input-suffix"
    ><slot name="suffix"></slot
  ></span>
</div>
I suggest for the span, onFocus redirects the focus to the input.

**Expected behavior**
A user should be able to click within the visual border of the input and begin to type
**Screenshots**

**CodeSandbox**
If applicable, please provide a CodeSandbox or other online code editor tool link that contains an example of the bug.
in addition to the gif above
can't input text with cursor here here:
![image](https://github.com/user-attachments/assets/995d02e7-98af-45fb-9790-5b73778ac6fe)
can input text with cursor here
![image](https://github.com/user-attachments/assets/771a60fe-10c7-498f-867f-62fd582dff10)

**Environment (please complete the following information):**

Chrome Version 128.0.6613.138 (Official Build) (64-bit)
issue present here: https://astro-components.netlify.app/?path=/docs/forms-input--invalid
and in our angular app

**Additional context**

kiley-mitti:
Hi @danielwlemmon thank you for bringing this to our attention! I think I understand what is happening here, but can you confirm something for me? Is the 'Navigate pages with a text cursor' setting enabled in Chrome? To find out, in Chrome go to Settings > Accessibility and it should be the 4th option down. (Just want to make sure that I'm recreating the issue properly.)
Thanks!

danielwlemmon:
yes, it is enabled
![image](https://github.com/user-attachments/assets/7339660a-9b2c-451e-84ee-57f9ef37debd)

[...]

PR #1341, opened 2024-09-17
Fix caret navigation inside rux-input

kiley-mitti:
## Brief Description

When the accessibility option for Caret Navigation is turned on in a browser, the text navigation falls inside the rux-inut spans but outside the actual input box, making it appear as though the user can type into the input when they cannot.

This is a simple fix which turns off that navigation for the exterior div of the input and then reenables it in the input itself. I've tested this against normal inputs and have found that they now work similarly. You must still tab into an input to type into it, but the caret navigation no longer navigates to visually inside rux-input. 

## JIRA Link

[DIW-135](https://rocketcom.atlassian.net/jira/software/projects/DIW/issues/DIW-135)

## Motivation and Context

This issue was brought to our attention this morning. [#1340](https://github.com/RocketCommunicationsInc/astro/issues/1340)
This is meant to fix that oddity.

## Issues and Limitations

Full disclosure, this is the first time I've ever used caret nav. I've tested up against standard browser inputs and did some research on how it is supposed to function so I'm reasonably confident that this is the fix.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/index.html
micahjones13, line 1: Could you clear this index file? I think some of it is actually from me like 4 months ago 🫤

micahjones13 (approved):

kiley-mitti:
Thanks Micah :)

PR #1346, opened 2024-09-25
feat(push-button): add active and inactive label props

micahjones13:
## Brief Description

This PR adds `active-label` and `inactive-label` props to rux-push-button. This way it's easy to display a different label based on the push-buttons state. 
Adds a new 'Dynamic Label' story to Push Button.
This PR also updates Playwright to be 1.38. This means that I had to update a few tests in other components as well. 

## JIRA Link

https://rocketcom.atlassian.net/browse/DIW-140

## Motivation and Context

This is a pretty common use case and is doable without this change, but this makes it more convenient. 

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


packages/web-components/src/components/rux-push-button/rux-push-button.tsx
```diff
@@ -82,6 +91,7 @@ export class RuxPushButton {
 
     private _onChange(e: Event) {
         const target = e.target as HTMLInputElement
+        console.log('changing checked value')
```
FMorrison87, line 94: Remove log :)

packages/web-components/src/components/rux-push-button/rux-push-button.tsx
```diff
@@ -90,11 +100,21 @@ export class RuxPushButton {
         this.ruxBlur.emit()
     }
 
+    private _determineLabelText(): string {
+        console.log(`checked: ${this.checked}`)
```
FMorrison87, line 104: Remove log :)

micahjones13:
There will be a layout shift based on the label if the active/inactive labels are different lengths. I believe this could be compensated for by setting the width to accommodate the longer label text by default. Moving this to a draft PR while I figure that out.

PR #1349, opened 2024-09-27
feat(tab + Storybook): tab action slot and Storybook upgrade

FMorrison87:
## Brief Description

This PR includes a relatively small change to `rux-tab` to add an `actions` slot. The bigger change is upgrading Storybook to ~8, which included modifying every single story in the project, as well as the controlling Storybook files to accommodate the major breaking changes of that upgrade.

## JIRA Link

None from a public-facing JIRA.

## Motivation and Context

The motivation for adding and actions slot to the tab component was to mitigate a problem from the implementation of a common pattern that we see in tabs used in space applications: closable tabs. As tabs used to be configured, any additional button or icon on the tab itself would fire off the `ruxselected` event which handles the selecting of the tabs, however this is problematic when you're trying to close a tab without selecting it. This problem required hacky work-arounds in every scenario. The solution of adding in a new tab also implements that tab not firing off `rux-selected`, thus allowing the developer to execute actions on a tab without that tab being automatically selected.

The second part of this PR was to upgrade Storybook from ~6 --> ~8. This was a massive upgrade and included major breaking changes to the way stories were configured and written. Each story for every component was modified to suit the new Storybook. This upgrade needed to be undertaken at some point so that we could continue using Storybook for our component documentation.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


kiley-mitti (approved):
no notes

PR #1363, opened 2025-01-10
Mj.179 timeline legend

micahjones13:
## Brief Description

Adds several new features to rux-timeline: 
- `ruler-position` prop which allows the user to control the placement of the ruler, either top, bottom or both. Defaults to both as per design. 
- `show-grid` prop displays a grid around intervals, making the timeline easier to parse. (Shout out to @kiley-mitti and @FMorrison87 for making this) 
- `show-secondary-ruler` adds the date to the ruler as well as the time. This also displays a Julian Day by default, per design. 
- `hide-j-day` prop toggles the display of the Julian day shown when `show-secondary-ruler` is set to true. 

The overall style of the timeline has slightly changed, mostly revolving around the ruler portion. 
Added tests and stories for new props. 

### Deprecation of `show-start-of-day`
The `show-start-of-day` functionality is obsolete now that we have the `show-secondary-ruler` which shows intervals in a cleaner way. I've added deprecation warnings of this property to Storybook, as well as a console warning when using `show-start-of-day`. 

## JIRA Link

https://rocketcom.atlassian.net/browse/R2C2-179

## Motivation and Context

Update rux-timeline to better align with the feel and style of upcoming vis-timeline version. 

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


.github/workflows/pull-request.yml
```diff
@@ -13,7 +13,7 @@ env:
 jobs:
   setup:
     if: github.event.pull_request.draft == false
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
micahjones13, line 16: Ubutntu latest (24.04) has an issue with some packages, so the github actions were never installing. I downgraded for the time being.

FMorrison87: Good catch!

packages/web-components/playwright.config.ts
```diff
@@ -31,10 +32,10 @@ const config: PlaywrightTestConfig<TestOptions> = {
              * Increases the maximum allowed pixel difference to account
              * for slight browser rendering inconsistencies.
              */
-            // maxDiffPixelRatio: 0.01,
+            maxDiffPixelRatio: 0.02,
```
micahjones13, line 35: I've updated these because rux-clock's time would change during the VRT's, and thus they'd fail by 0.01% of pixels being off.

FMorrison87 (commented):
This is some amazing work. The comments I had were just nits. If you want to meet next week to go through all of the math you had to write I'd be happy to do that. Honestly I probably would need you to walk me through it anyway.

Nice work on this! This one was a doozy.

packages/angular-workspace/projects/angular/src/directives/proxies-list.ts
```diff
@@ -1,4 +1,3 @@
-// @ts-nocheck
```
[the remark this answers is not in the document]
micahjones13: A long time ago in a place far far away that file used to throw a TS error so I wrote a middleware that simply added that ignore. I'm honestly not sure why the tsignore got removed here, but it seems to work so maybe we don't need the middleware anymore. I'll check.

FMorrison87 (approved):
This looks great! Nice work! 🔥

PR #1370, opened 2025-03-05
feat: add datetime-picker component

micahjones13:
## Brief Description

Adds the `rux-datetime-picker` component. This component uses the new `rux-calendar` and `rux-day` components. 

## General Notes

Props

`<rux-datetime-picker>`
- `precision`: Controls the granularity of the time (hour, min, sec, ms)
- `julian-format`: Controls whether or not the datepicker is in julian mode
- `max-year`: The max year value the datepicker can have 
- `min-year`: The min year value the datepicker can have 
- `help-text`: The help text. Same as our `rux-input`
- `error-text`: The error text. Same as our `rux-input`
- `disabled`: renders the datepicker as disabled. 
- `label`: Controls the label text for the datepicker 
- `name`: Controls the name of the datepicker input
- `value`: The value of the datepicker. Must be a valid ISO or Julian ISO string
- `size`: The size of the datepicker. small, medium or large 

`<rux-calendar>` and `<rux-day>` both have their own props which are passed down the component tree originating from `<rux-datetime-picker>`. These props are internal for the time being, until the need for a standalone calendar arises. 

Events
- `<rux-datetime-picker>`
    - rux-change: fired when the value of rux-datetime-picker is changed and commited.
        - This also fires when a rux-day is selceted via the calendar. 
    - rux-input: fires when the user inputs values into the datepicker 
    - rux-blur: fires when datetime-picker loses focus 

## Motivation and Context

Adds an astro component for handling datetime selection. Provides users with a julian formatted datetime picker. 

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist


Paterma:
https://github.com/user-attachments/assets/bc125deb-7118-436f-ae75-791dfed7628a

Awesome job so far! The biggest thing I've noticed is the time won't update unless the date val is changed

Paterma:
Functionality wise - couldn't find a thing wrong when testing them all! Very impressed 👏🏼 
You were missing the `precision="min"` prop on the Minutes datepicker under precision examples but it's obviously not a bug - just pointing it since it defaults to milliseconds. Code looks solid, too. Aside from loading it up to storybook - I think you're good!

packages/web-components/src/components/rux-datetime-picker/rux-datetime-picker.tsx
```diff
@@ -0,0 +1,1077 @@
...
+        this.ruxDatetimePickerChange.emit(
+            this.parts.find((part) => part.type === 'day')?.value
+        )
+    }
+
+    toggleCalendar() {
+        this.isCalendarOpen = !this.isCalendarOpen
+    }
+
+    determineMinMax(type: PartKey, isMicro: boolean, isJulian?: boolean) {
```
brothhammer, line 810: This is a very expensive calculation in the render method. We should look into a way to memoize in Stencil. This method is called multiple times per render cycle but the result only depends on minYear, maxYear and JulianFormat - that don't change often. We want to avoid frequent recalculations of the same min/max values.

packages/web-components/src/components/rux-datetime-picker/rux-datetime-picker.tsx
```diff
@@ -0,0 +1,1077 @@
...
+                                        key={i}
+                                    >
+                                        {value}
+                                    </span>
+                                ) : (
+                                    <Fragment>
+                                        <input
+                                            key={i}
+                                            class={{
+                                                part: true,
```
brothhammer, line 948: This class object is in the render method and recreated with each render. It only needs to be recreated if type or precision changes.

packages/web-components/src/components/rux-datetime-picker/rux-datetime-picker.tsx
```diff
@@ -0,0 +1,1077 @@
...
+                    const currentYear = new Date().getUTCFullYear()
+                    value = `${currentYear}-${value}`
+                }
+
+                // --- MICROSECOND HANDLING ---
+                if (isMicro && !this.julianFormat) {
+                    //check if incoming value is in an Oridnal ISO format. If so, convert it to gregorian since we're not in julianFormat.
+                    const ordinalFormatMatch = value.match(
+                        /^([0-9]{4})(?:-([0-9]{1,3}))?(?:T([0-9]{2})(?::([0-9]{2}))?(?::([0-9]{2})(?:\.([0-9]{1,6}))?)?Z?)?$/
+                    )
```
brothhammer, line 311: Regex compilation is expensive and this is static. We should move this so it doesn't need to get recompiled for every call of `handleInitialValue`

brothhammer (changes requested):
Wow! this is really impressive and myself coming in at the last minute is struggling to understand how you have kept everything straight for so long. 

I have only dug into rux-datetime-picker.tsx. There are places where we are making fairly expensive calculations in the render method that would be good to memoize or in some cases move to a constants file (like the static regex). I also took a shot at code splitting this file on a seperate branch https://github.com/RocketCommunicationsInc/astro/pull/1408

For maintainability our future selves and other developers I think we need to continue to explore ways to break up rux-datetime-picker.tsx and rux-calendar.tsx. In my view it is okay to have very long files when that file is doing the same thing over and over again (like a testing file or routing file). I'm not married to how I split things and was making some pretty quick decisions. 

Additionally given the size of this PR I would greatly appreciate a code walk through meeting to better understand.

Again this is very impressive and I believe the functionality is there. Just want to move it in a direction that can be more easily understood. Happy to pair up or look further into how memoization works in Stencil to keep this moving.

PR #1403, opened 2025-06-19
Mj.tab style

micahjones13:
## Brief Description

Adds a `compact` prop to `rux-tabs` that allows for the [compact tab styling](https://www.figma.com/design/9gMquz9RpfIfkK4V2HZjic/New-Tab-Bars?node-id=1-328&t=pjTL47o0s5Iw114D-0).

## Motivation and Context

Update tabs to match designs in community file.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

- [x] This PR adds or removes a Storybook story.
- [ ] I have added tests to cover my changes.
- [ ] Regressions are passing and/or failures are documented
- [ ] Changes have been checked in evergreen browsers

PR #1417, opened 2025-07-27
chore: remove console.log

snyamathi:
Looks like this snuck in during https://github.com/RocketCommunicationsInc/astro/pull/1403 cc @micahjones13


micahjones13 (approved):

micahjones13:
Thanks so much @snyamathi!

PR #1500, opened 2026-04-28
Angular form element accessor bugfix

NicDominguez:
## Types of changes

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


packages/angular-workspace/projects/angular/src/lib/angular-lib.module.ts
```diff
@@ -3,17 +3,17 @@ import { CommonModule } from '@angular/common';
...
 defineCustomElements(window);
 
+const VALUE_ACCESSORS = [BooleanValueAccessor, NumericValueAccessor, TextValueAccessor];
+
 @NgModule({
-  declarations: [],
+  declarations: [...DIRECTIVES, ...VALUE_ACCESSORS],
   imports: [CommonModule],
-  exports: [],
+  exports: [...DIRECTIVES, ...VALUE_ACCESSORS],
```
nortonprojects: It's definitely necessary to get those directives exported from the module, but it'd also be great if they were included in the standalone components. Since you've got them all working properly, you could use [Host Directives](https://angular.dev/guide/directives/directive-composition-api) to ensure they're all working right as standalone components

PR #1526, opened 2026-08-05
fix: exported types for conditional exports

snyamathi:
## Brief Description

https://nodejs.org/api/packages.html#conditional-exports the `exports` takes precedence over the top level `types` field - as a result, typescript types are broken in v8 due to the changes in https://github.com/RocketCommunicationsInc/astro/commit/fecfc74198d16b145f42ceb67c5557d2fa10dda8#diff-7298512f791c54d2827facfc8d26ae8a5db2ac960e464e30953ccaab05dd412eR10

## Related Issue

https://github.com/RocketCommunicationsInc/astro/pull/1500 pretty sure the regression was introduced here

## General Notes

I've tested this locally using `patch-package` but I'm not sure how else to add a test for this

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/salty-showers-dig.md
+++ b/.changeset/salty-showers-dig.md
@@ -1,0 +1,5 @@
+---
+"root": patch
+---
+
+fix: exported types for conditional exports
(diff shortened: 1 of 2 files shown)


snyamathi:
Let me test this a bit more - not ready yet, will re-open.

